PR #7, opened 2018-04-05
Add refinement to trilinear elements

rchristie:


alan-wu:
Looking alright, I will merge this in now.

PR #13, opened 2018-06-14
Correctly simplify the installation process.

metatoaster:
Don't override the cmdclass for something that can be solved trivially with an additional line in requirements.txt.  Also document the command and specify the dependencies inside the setup function so that they will persist into the wheel.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,6 @@ SETUP_DIR = os.path.dirname(os.path.abspath(__file__))
 # List all of your Python package dependencies in the
 # requirements.txt file
 
-
 def readfile(filename, split=False):
     with io.open(filename, encoding="utf-8") as stream:
         if split:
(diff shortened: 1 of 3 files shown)


metatoaster:
Of course, if the packages are not on pypi (yet) the install command can be used, since downstream users won't see the requirements.txt, and if they (i.e. scaffoldmaker-web-demo) try to install it they will also need the contents there, but at least with this the missing dependencies will be flagged.

For instance

```
$ pip install -r requirements.txt
...
Obtaining scaffoldmaker from git+https://github.com/metatoaster/scaffoldmaker.git#egg=scaffoldmaker (from -r requirements.txt (line 1))
  git clone in ./.env35/src/scaffoldmaker exists with URL https://github.com/ABI-Software/scaffoldmaker.git
...
Collecting opencmiss.utils (from scaffoldmaker->-r requirements.txt (line 1))
  Could not find a version that satisfies the requirement opencmiss.utils (from scaffoldmaker->-r requirements.txt (line 1)) (from versions: )
No matching distribution found for opencmiss.utils (from scaffoldmaker->-r requirements.txt (line 1))
```

setup.py
```diff
@@ -18,19 +17,13 @@ def readfile(filename, split=False):
 readme = readfile("README.rst", split=True)[3:]  # skip title
 # For requirements not hosted on PyPi place listings
 # into the 'requirements.txt' file.
-requires = []  # minimal requirements listing
+requires = [
+    # minimal requirements listing
+    'opencmiss.utils',
+    'ZincPythonTools',
+]
```
alan-wu: We can also use the following instead 
```
['opencmiss.utils @ https://api.github.com/repos/OpenCMISS-Bindings/opencmiss.utils/tarball/master', 
'ZincPythonTools' @  https://api.github.com/repos/OpenCMISS-Bindings/ZincPythonTools/tarball/master']
```
This should work for downstream users such as scaffold-web-demo.

alan-wu:
I think this change should be accepted and I will be making some more changes as there are missing required packages such as numpy and scipy.

metatoaster:
Oh, now that you got this merged in with this modification, the requirements.txt can be removed and the installation instructions should be updated to `pip install -e .` in the README to further reduce typing.

alan-wu:
> Oh, now that you got this merged in with this modification, the requirements.txt can be removed and the installation instructions should be updated to `pip install -e .` in the README to further reduce typing.

Sure, I will update that now.

PR #21, opened 2018-09-30
Lens

mlin865:


scaffoldmaker/meshtypes/meshtype_3d_lens1.py
```diff
@@ -0,0 +1,250 @@
...
+                options[key] = 0.0
+        if options['Spherical radius fraction'] > 1.0:
+            options['Spherical radius fraction'] = 1.0
+        if options['Posterior radius of curvature'] < options['Axial thickness']*0.5:
+            options['Posterior radius of curvature'] = options['Axial thickness']*0.5
+        if options['Anterior radius of curvature'] < options['Axial thickness']*0.5:
+            options['Anterior radius of curvature'] = options['Axial thickness']*0.5
+
+    @classmethod
+    def sphereToLens(cls, region, options):
```
rchristie: Pass actual parameters to this function, not options dict. I.e.. coordinates, radiusSphere, radiusAnt, radiusPos, lensThickness, sphericalRadiusFraction etc. Describe each argument using :param NAME: description.

Can get fieldmodule from coordinates field, so no need to pass:
fm = coordinates.getFieldmodule()

Function and description should state that it defines a field giving positions on a spherical lens as functions of positions in the sphere.

I would make this a standalone function e.g.:
def getSphereToLensCoordinates(args):

scaffoldmaker/meshtypes/meshtype_3d_lens1.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        Map coordinates of the sphere within a width limit olim to arcs with radii 
+        rAnt and rPos, respectively. Elements maintain constant size radially and match
+        up at dlim. Outside of rlim, affine transformation applied to transform sphere 
+        coordinates to be tangential to morphed spherical surfaces
+        return: lensRC
+        """
+        fm = region.getFieldmodule()
+        fm.beginChange()
+        cache = fm.createFieldcache()
```
rchristie: This is unused. Remove.

scaffoldmaker/meshtypes/meshtype_3d_lens1.py
```diff
@@ -0,0 +1,250 @@
...
+from scaffoldmaker.meshtypes.meshtype_3d_solidsphere1 import MeshType_3d_solidsphere1
+from scaffoldmaker.utils.meshrefinement import MeshRefinement
+from scaffoldmaker.utils.zinc_utils import *
+from opencmiss.zinc.element import Element, Elementbasis, Elementfieldtemplate
+from opencmiss.zinc.field import Field
+from opencmiss.zinc.node import Node
+
+class MeshType_3d_lens1:
+    '''
+    classdocs
```
rchristie: Describe class here. Can be same text as at top of file, but more detailed.
Worth mentioning that it is a spherical lens, and that the lens is created by a function morphing a sphere to the lens.

scaffoldmaker/meshtypes/meshtype_3d_lens1.py
```diff
@@ -0,0 +1,250 @@
...
+            'Anterior radius of curvature',
+            'Axial thickness']:
+            optionNames.insert(3, optionName)
+        return optionNames
+
+    @staticmethod
+    def checkOptions(options):
+        MeshType_3d_solidsphere1.checkOptions(options)
+        for key in [
+            'Spherical radius fraction',
```
rchristie: Remove Posterior/Anterior radii from minimum zero as handled below.
I like the check options to be in the order they appear in the dictionary (except where a the same check is applied to several options e.g. if options[key] < 0.0.
I.e. check min 0.0 for 'Axial thickness', 'Spherical radius fraction'. Then min Anterior radius, min Posterior radius. Use 'for key in [anterior/posterior]' as same expression. Lastly limit Spherical radius fraction to max 1.0.

PR #39, opened 2019-02-14
New MeshTypes + vector + transformation

mahyar-osn:
I have added two new MeshTypes: 1) Elliptic Cylinder 1 and 2) Vertebra 1.
These two are (for now) the same with only some small differences in their parameters. The idea is to work on the vertebra 1 mesh until it looks like one.

In addition, I have added three new methods in the vector.py to:
- get the diagonal of a matrix,
- generate an identity matrix, and
- compute the outer product of two vectors.

Furthermore, as discussed, I have created a brand new file called transformation.py where we could put any transformation-related functions there. Currently, there is only one method which returns a rotation matrix rotating around an axis.


scaffoldmaker/meshtypes/meshtype_3d_ellipticcylinder1.py
```diff
@@ -0,0 +1,476 @@
...
+        height = options['Height']
+        useCubicHermiteThroughWall = True
+
+        a = identity(4)
+
+        zero = [0.0, 0.0, 0.0]
+
+        """ Get nodes along the centre line """
+        cx = [zero, [0.0, 0.0, height]]
+        cd1 = [[0.0, 0.0, height / elementsCountUp], [0.0, 0.0, height / elementsCountUp]]
```
rchristie: Don't need to divide by elementsCountUp here for sampleCubicHermiteCurves to work as expected.

scaffoldmaker/meshtypes/meshtype_3d_ellipticcylinder1.py
```diff
@@ -0,0 +1,476 @@
...
+import math
+
+from scaffoldmaker.meshtypes.scaffold_base import Scaffold_base
+from scaffoldmaker.utils.meshrefinement import MeshRefinement
+from scaffoldmaker.utils.eftfactory_bicubichermitelinear import eftfactory_bicubichermitelinear
+from scaffoldmaker.utils.eftfactory_tricubichermite import eftfactory_tricubichermite
+from scaffoldmaker.utils.zinc_utils import *
+from scaffoldmaker.utils.geometry import *
+from scaffoldmaker.utils.interpolation import *
+from scaffoldmaker.utils.vector import *
```
rchristie: The function names in vector are so generic I prefer this to be imported with:
import scaffoldmaker.utils.vector as vector
& all functions called with vector.magnitude etc.

scaffoldmaker/meshtypes/meshtype_3d_ellipticcylinder1.py
```diff
@@ -0,0 +1,476 @@
...
+    """
+
+    @staticmethod
+    def getName():
+        return '3D Elliptic Cylinder 1'
+
+    @staticmethod
+    def getDefaultOptions(parameterSetName='Default'):
+        return {'Number of elements around': 4, 'Number of elements up': 3, 'Number of elements through wall': 2,
+                'Major diameter': 1.2, 'Minor diameter': 0.7, 'Height': 2.0, 'Thickness ratio': 0.5,
```
rchristie: 'Thickness ratio' is too vague. Qualify with other words e.g. Longitudinal & what it's a ratio of end to middle?

scaffoldmaker/meshtypes/meshtype_3d_ellipticcylinder1.py
```diff
@@ -0,0 +1,476 @@
...
+
+        zero = [0.0, 0.0, 0.0]
+
+        """ Get nodes along the centre line """
+        cx = [zero, [0.0, 0.0, height]]
+        cd1 = [[0.0, 0.0, height / elementsCountUp], [0.0, 0.0, height / elementsCountUp]]
+        cd2 = [[0.0, axisB / elementsCountThroughWall, 0.0], [0.0, axisB / elementsCountThroughWall, 0.0]]
+        cd3 = [[axisA / elementsCountThroughWall, 0.0, 0.0], [axisA / elementsCountThroughWall, 0.0, 0.0]]
+
+        t = options['Thickness ratio']
```
rchristie: Do you expect this to vary along the cylinder? If not these are just constant.

scaffoldmaker/meshtypes/meshtype_3d_ellipticcylinder1.py
rchristie: Also, we've traditionally made xi1 go around meshes, xi2 go along meshes and xi3 go radially. In the other scaffolds this is followed, and s1, s2 and s3 do the same thing. This would mean your cd1 should be cd2, etc. Have a look at the 3d tube.

This may eventually change for long tube meshes such as the colon as we probably want nodes varying slowest along the tube as it's the most obvious partition for an efficient matrix.

scaffoldmaker/meshtypes/meshtype_3d_ellipticcylinder1.py
```diff
@@ -0,0 +1,476 @@
...
+        else:
+            prevBinormal = crossproduct3(prevUnitTangent, [0.0, -1.0, 0.0])
+        prevUnitBinormal = normalise(prevBinormal)
+        prevUnitNormal = crossproduct3(prevUnitBinormal, prevUnitTangent)
+        sNormal.append(prevUnitNormal)
+        sBinormal.append(prevUnitBinormal)
+
+        # Step through central line and rotate central line axes to align tangent
+        # to tangent from previous level
+        for n in range(1, elementsCountUp + 1):
```
rchristie: I'm not sure what this is achieving and why the whole rotation matrix stuff is needed. We will need to discuss.

scaffoldmaker/utils/transformation.py
```diff
@@ -0,0 +1,35 @@
...
+
+@author: Mahyar Osanlouy
+"""
+
+
+import math
+
+from scaffoldmaker.utils.vector import *
+
+def rotationMatrix(angle, direction, point=None):
```
rchristie: createRotationMatrixFromAngleAxis? There are several ways to make rotation matrices.
Also, I don't see how you can use point as you'll need a separate translation vector or use 4x4 homogeneous transformations.

scaffoldmaker/utils/vector.py
```diff
@@ -38,3 +39,34 @@ def setMagnitude(v, mag):
...
     return [ c*scale for c in v ]
+
+def getDiagonal(m):
+    '''
+    :param m: matrix
+    :return: Diagonal elements of m in a list.
+    '''
+    width, height = len(m[0]), len(m)
+
+    def diag(sx, sy):
```
rchristie: These seven lines of code seem pretty opaque to me. I haven't time to deeply analyse them but in particular I can't see how any of the later return statements are ever reached?

scaffoldmaker/utils/vector.py
```diff
@@ -38,3 +39,34 @@ def setMagnitude(v, mag):
     '''
     scale = mag/math.sqrt(sum(c*c for c in v))
     return [ c*scale for c in v ]
+
+def getDiagonal(m):
```
rchristie: This is used as follows:
R = getDiagonal([cosa, cosa, cosa])
Does that mean it should be renamed createDiagonalMatrix?
Creates a square matrix with the supplied values on its diagonal and zeros everywhere else.
:param d: List of values on diagonal.
:return: Square matrix as list (columns) of lists (rows) of values.

mahyar-osn:
Closing this PR. A new PR will be done soon.

PR #43, opened 2019-03-04
Colon

mlin865:


scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -0,0 +1,448 @@
...
+        refineElementsCountAlong = options['Refine number of elements along']
+        refineElementsCountThroughWall = options['Refine number of elements through wall']
+
+        baseRegion = region.createRegion()
+        cls.generateBaseMesh(baseRegion, options)
+
+        meshrefinement = MeshRefinement(baseRegion, region)
+        meshrefinement.refineAllElementsCubeStandard3d(refineElementsCountAround, refineElementsCountAlong, refineElementsCountThroughWall)
+
+def getUnitHaustraOuter(elementsCountAround, elementsCountAlongHaustra, radius, cornerRadiusFraction,
```
rchristie: Am I right that this makes a segment of the colon consisting of the full circumference with 3 haustra around, one along?
Rename getColonHaustraSegmentOuterPoints. Unit as an adjective (before) means unit sized to me.
Could alternatively generate inner points.

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -0,0 +1,448 @@
...
+    :param haustraRadiusFraction: Factor is multiplied by inner haustra
+    radius to obtain radius of intersecting circles in the middle cross-section
+    along a haustra.
+    :param foldFactor: Factor is multiplied by haustra length to scale derivative
+    along the haustra at the interhaustral septa end.
+    :param haustraCurvatureFactor: Factor is multiplied by haustra length to 
+    scale derivative along the haustra at the middle section of the haustra.
+    :param wallThickness: Thickness of haustra through wall.
+    :param haustraLength: Length of a haustra unit.
+    :return: coordinates, derivatives on outer surface of haustra unit.
```
rchristie: :return: xOuter, d1Outer, d2Outer, unitZ
What is unitZ needed for? Can easily get from cross product?

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -0,0 +1,448 @@
...
+
+    dx_ds1InnerList = dx_ds1InnerList + d1Inner
+
+    for n in range(len(xInnerList)):
+        dx_ds3 = crossproduct3(normalise(dx_ds1InnerList[n]), normalise(dx_ds2InnerList[n]))
+        unitdx_ds3 = normalise(dx_ds3)
+        dx_ds3InnerUnitList.append(unitdx_ds3)
+
+    # Pre-calculate node locations and derivatives on outer boundary
+    xOuter, d1Outer, d2Outer = getOuterCoordinatesAndDerivativesFromInner(xInnerList, dx_ds1InnerList, dx_ds2InnerList, dx_ds3InnerUnitList, wallThickness, elementsCountAlongHaustra, elementsCountAround)
```
rchristie: If it's easier to generate the inner points, do so.

scaffoldmaker/meshtypes/meshtype_3d_haustra1.py
```diff
@@ -0,0 +1,511 @@
+"""
+Generates a 3-D haustra mesh along a central line, with variable
```
rchristie: I wasn't expecting this to be a separate scaffold, but it could be useful.
Does it make a straight 'colon haustra segment' such as would be repeated on the colon? (See earlier comment).
If so I'd expect the colon to inherit most of its parameters from this scaffold, and use a common function in this scaffold for getting the outer points. Both can use the tube mesh (new version), but this one creates its own single element centre line of the same length as the haustrum.

scaffoldmaker/utils/tubemesh2.py
```diff
@@ -0,0 +1,356 @@
...
+    :param wallThickness: thickness of wall
+    :param unitAxis: axis of unit profile
+    :param unitLength: length of unit profile
+    :param useCubicHermiteThroughWall: use linear when false
+    '''
+    zero  = [0.0, 0.0, 0.0]
+    elementsCountAlong = elementsCountAlongUnit*unitsCountAlong
+
+    # Sample central line to get same number of elements as elementsCountAlong
+    sx, sd1, se, sxi, _ = sampleCubicHermiteCurves(cx, cd1, elementsCountAlong)
```
rchristie: As this is not using arcLengthDerivatives=true, it won't match the length returned for the colon, so haustra will not finish exactly at the end.

scaffoldmaker/utils/tubemesh2.py
```diff
@@ -0,0 +1,356 @@
+'''
+Utility function for generating tubular mesh from a central line
+using a unit profile.
+Created on 18 Feb, 2019
+
+@author: Mabelle Lin
```
rchristie: We're now removing all @author throughout the code (including my own); the record of who did what is in git.

scaffoldmaker/utils/tubemesh2.py
```diff
@@ -0,0 +1,356 @@
+'''
```
rchristie: Merge this new implementation into tubemesh.py, replacing old one.
The original centrallinetube1 scaffold can be migrated or discarded. I'm guessing what it brought to the table was ellipse profiles with equal element spacing, so not sure how/whether to support?
What you have is likely able to get into the code for now, so following are largely future discussions:
Soon we will need variable wall thickness around and along units/segments, plus in each layer, and also controllable side axis or axes to produce twists. Here I'm thinking your approach of having unit side axes and having all sizing information in the unit/segments could work well, then e.g. ellipsoidal sections can be passed in.
Furthermore, eventually the nodes should be numbered from one end to the other, and elements too. However, xi1 can remain around, xi2 along and xi3 through the wall.
Same function should also support generating 2D meshes in future.

scaffoldmaker/utils/tubemesh2.py
```diff
@@ -0,0 +1,356 @@
...
+            factor = 1.0 - curvatureAround*thickness*xi3
+            dx_ds2 = [ factor*c for c in d2Outer[n]]
+            dx_ds2List.append(dx_ds2)
+            #dx_ds3
+            dx_ds3 = [c * thickness/elementsCountThroughWall for c in norm]
+            dx_ds3List.append(dx_ds3)
+
+    return xList, dx_ds1List, dx_ds2List, dx_ds3List
+
+def rotationMatrixAboutAxis(rotAxis, theta):
```
rchristie: See earlier comment.
Whenever you find you've put the same function in 2 files you should remove it to a common file.

scaffoldmaker/utils/tubemesh2.py
```diff
@@ -0,0 +1,356 @@
...
+from scaffoldmaker.utils.eftfactory_tricubichermite import eftfactory_tricubichermite
+from scaffoldmaker.utils.zinc_utils import *
+from scaffoldmaker.utils.geometry import *
+from scaffoldmaker.utils.interpolation import *
+from scaffoldmaker.utils.vector import *
+from opencmiss.zinc.element import Element, Elementbasis
+from opencmiss.zinc.field import Field
+from opencmiss.zinc.node import Node
+
+def generatetubemesh(region,
```
rchristie: Note I haven't studied the implementation of this function, just the interface (which is fine for now).
I'm happy to bring it in as-is now, but will look at it in more detail when additional complications for variable wall thickness are added.

mlin865:
Cancel this pull request in view of revisions made on a new branch with a new pull request.

PR #46, opened 2019-03-22
Texture coordinates

mlin865:


scaffoldmaker/utils/tubemesh.py
```diff
@@ -243,6 +243,107 @@ def generatetubemesh(region,
...
+                    1.0 / elementsCountThroughWall * n3]
+                d1 = [1.0 / elementsCountAround, 0.0, 0.0]
+                d2 = [0.0, 1.0 / elementsCountAlong, 0.0]
+                uTexture.append(u)
+                d1Texture.append(d1)
+                d2Texture.append(d2)
+
+    nodeIdentifier = nextNodeIdentifier
+    for n in range(len(uTexture)):
+        if n%(elementsCountAround+1) == 0.0:
```
rchristie: == 0, not a floating point.
Also, I like spaces around operators % and +.
That said, I would loop as follows:
```
for n3 in range(elementsCountThroughWall + 1):
    for n2 in range(elementsCountAlong + 1):
        for n1 in range(elementsCountAround):
```
... and then compute coordinates from those precalculated around one slice.

PR #58, opened 2019-08-06
Serialise bytes string to JSON

rchristie:
Zinc now returns a bytes object when fetching serialisation data from a streamresource memory. Now decoding to utf-8 to allow serialisation to JSON.
Also fixed error when calling super class in encoder.


alan-wu:
Thanks, looks good.

PR #59, opened 2019-09-27
Vary inner radius and tenia coli width along colon length

mlin865:


scaffoldmaker/utils/tubemesh.py
```diff
@@ -20,13 +20,14 @@ def generatetubemesh(region,
     elementsCountThroughWall,
     segmentCountAlong,
     cx, cd1, cd2, cd12,
-    xInner, d1Inner, d2Inner, wallThickness,
-    segmentAxis, segmentLength,
+    innerRadiusList, dInnerRadiusList,
+    tcWidthList, dTCWidthList,
```
rchristie: Can't pass colon-specific parameters in this way as it takes away the generic-ness of tubemesh.
All varying properties need to be passed to the segment generator, including radius and TC width. This makes sense as they are specific to that generator.
Tubemesh will supply distance along when it calls the segment generator which can use it to sample the varying properties.

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -302,28 +395,38 @@ def generateBaseMesh(region, options):
...
+                segmentLength, wallThickness, cornerInnerRadiusFactor, haustrumInnerRadiusFactor)
 
         # Generate tube mesh
-        annotationGroups, nextNodeIdentifier, nextElementIdentifier, xList, d1List, d2List, d3List, sx, curvatureAlong, factorList = tubemesh.generatetubemesh(region,
-            elementsCountAround, elementsCountAlongSegment, elementsCountThroughWall, segmentCount, cx, cd1, cd2, cd12,
-            xInner, d1Inner, d2Inner, wallThickness, segmentAxis, segmentLength, useCrossDerivatives, useCubicHermiteThroughWall,
-            annotationGroups, annotationArray, transitElementList, uList, arcLengthOuterMidLength)
+        annotationGroups, nextNodeIdentifier, nextElementIdentifier, xList, d1List, d2List, d3List, sx, curvatureAlong, factorList, uList, relaxedLengthList, tubeTCWidthList = tubemesh.generatetubemesh(region,
+           elementsCountAround, elementsCountAlongSegment, elementsCountThroughWall, segmentCount, cx, cd1, cd2, cd12, innerRadiusSegmentList, dInnerRadiusSegmentList,
+           tcWidthSegmentList, dTCWidthSegmentList, tubeMeshSegmentInnerPoints, wallThickness, segmentLength, useCrossDerivatives, useCubicHermiteThroughWall)
```
rchristie: Can't pass TC width here - see comments on tubemesh class.

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -350,3 +453,59 @@ def generateMesh(cls, region, options):
...
+    nx = []
+    nd1 = []
+    xList = []
+    d1List = []
+    for i in range(len(lengthList)):
+        v = [lengthList[i], paramList[i], 0.0]
+        d1 = [(lengthList[i+1] - lengthList[i]) if i < len(lengthList)-1 else (lengthList[i] - lengthList[i-1]), 0.0, 0.0]
+        nx.append(v)
+        nd1.append(d1)
+    smoothd1 = interp.smoothCubicHermiteDerivativesLine(nx, nd1)
```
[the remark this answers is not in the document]
mlin865: Discussed offline with Richard. Decided to keep things as they are now and will modify later when requirements change.

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -350,3 +453,59 @@ def generateMesh(cls, region, options):
...
+
+    # Get total arclength
+    arcLength = 0.0
+    for e in range(len(nx)-1):
+        arcLength += interp.getCubicHermiteArcLength(nx[e], smoothd1[e], nx[e + 1], smoothd1[e + 1])
+
+    # Find parameter value at each segment
+    xTol = 1.0E-6
+    for n in range(segmentCount + 1):
```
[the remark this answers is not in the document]
mlin865: interp.sampleCubicHermiteCurves will sample the curve describing how parameter changes along the centre line into equal arclength, but here I want to sample along the center line and get the parameter value along each of these points on the center line.

mlin865:
I've made the changes as per request, please review. Thanks!

scaffoldmaker/utils/interpolation.py
```diff
@@ -787,3 +787,61 @@ def getDoubleCubicHermiteCurvesMidDerivative(ax, ad1, mx, bx, bd1):
     magb = vector.magnitude(bd1)
     magm = arcLengtha + arcLengthb - 0.5*(maga + magb)
     return vector.setMagnitude(md1, magm)
+
+def sampleParameterAlongCenterLine(paramList, lengthList, outputElementsCount):
```
rchristie: Remove 'Center' from name as redundant (I try to avoid using names where there are US spelling differences anyway). I would pass the lengthList as first argument because it is the reference onto which the parameters are mapped (the x to the y).

scaffoldmaker/utils/interpolation.py
```diff
@@ -787,3 +787,61 @@ def getDoubleCubicHermiteCurvesMidDerivative(ax, ad1, mx, bx, bd1):
...
+    xList = []
+    d1List = []
+    for i in range(len(lengthList)):
+        v = [lengthList[i], paramList[i]]
+        d1 = [(lengthList[i+1] - lengthList[i]) if i < len(lengthList)-1 else (lengthList[i] - lengthList[i-1]), 0.0]
+        nx.append(v)
+        nd1.append(d1)
+    smoothd1 = smoothCubicHermiteDerivativesLine(nx, nd1)
+
+    # Get total arclength
```
rchristie: I think this should be sampling at equal distances along the lengths, not including the parameter as a component of coordinates (it may after all be used for a non-spatial value). It would look increasingly wrong if there was a lot of variation of the parameter at one end vs. the other.
I think we need a special smoothing function to get derivatives of parameter p with linearly varying x lengths.
The arc length should just be the total length of the line (last x - first x). Find xi in the original elements at sample lengths, then evaluate p, dp/dxi, dx/dxi and use chain rule to get dp/dx. Divide by outputElementsCount to get derivatives for the segment length?
Probably easier to describe on paper.

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -302,30 +397,109 @@ def generateBaseMesh(region, options):
...
+        d1Extrude = []
+        d2Extrude = []
+        d3UnitExtrude = []
+
+        # Create object
+        tubeMeshSegmentInnerPoints = TubeMeshSegmentInnerPoints(
+            region, elementsCountAroundTC, elementsCountAroundHaustrum, elementsCountAlongSegment,
+            tcCount, segmentLengthEndDerivativeFactor, segmentLengthMidDerivativeFactor,
+            segmentLength, wallThickness, cornerInnerRadiusFactor, haustrumInnerRadiusFactor,
+            innerRadiusSegmentList, dInnerRadiusSegmentList, tcWidthSegmentList, dTCWidthSegmentList)
```
rchristie: This object has a generic name but takes TC arguments?

scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -6,13 +6,13 @@
 
 import copy
 from scaffoldmaker.meshtypes.meshtype_1d_path1 import MeshType_1d_path1, extractPathParametersFromRegion
-from scaffoldmaker.meshtypes.meshtype_3d_colonsegmentsimplemesentery1 import MeshType_3d_colonsegmentsimplemesentery1, getColonSegmentInnerPointsNoTeniaColi
-from scaffoldmaker.meshtypes.meshtype_3d_colonsegmentteniacoli1 import MeshType_3d_colonsegmentteniacoli1, getColonSegmentInnerPointsTeniaColi, getTeniaColi
+from scaffoldmaker.meshtypes.meshtype_3d_colonsegment1 import MeshType_3d_colonsegment1, TubeMeshSegmentInnerPoints, getTeniaColi, createFlatAndTextureCoordinatesTeniaColi, createNodesAndElementsTeniaColi
```
rchristie: TubeMeshSegmentInnerPoints name is too generic; should be renamed to describe that it's for colon only. Probably want to rename the original class in this case, but worth knowing you can always rename something on import:
from scaffoldmaker.meshtypes.meshtype_3d_colonsegment1 import TubeMeshSegmentInnerPoints as ColonSegmentTubeMeshSegmentInnerPoints
This is often needed to disambiguate objects with the same name in two sources.

PR #60, opened 2019-11-08
Put source under src folder, add test

rchristie:


tests/test_heart.py
```diff
@@ -0,0 +1,58 @@
...
+from scaffoldmaker.utils.zinc_utils import evaluateFieldRange
+
+def assertAlmostEqualList(testcase, actualList, expectedList, delta):
+    assert len(actualList) == len(expectedList)
+    for actual, expected in zip(actualList, expectedList):
+        testcase.assertAlmostEqual(actual, expected, delta=delta)
+
+class HeartScaffoldTestCase(unittest.TestCase):
+
+    def test1_heart1(self):
```
hsorby: Test names should really start with `test_` this is what the test frameworks expect.

src/scaffoldmaker/utils/zinc_utils.py
```diff
@@ -4,10 +4,34 @@
...
+
+    def __enter__(self):
+        self._object.beginChange()
+        return self
+
+    def __exit__(self, *args):
+        self._object.endChange()
+
+
 def getOrCreateCoordinateField(fieldmodule, name='coordinates', componentsCount=3):
```
hsorby: These `getOrCreate*` functions could be canonicalised somewhat.

src/scaffoldmaker/utils/zinc_utils.py
```diff
@@ -269,6 +286,24 @@ def getMaximumElementIdentifier(mesh):
...
+        fieldcache = fieldmodule.createFieldcache()
+        result, minimums = fieldNodesetMinimum.evaluateReal(fieldcache, componentsCount)
+        assert result == RESULT_OK
+        result, maximums = fieldNodesetMaximum.evaluateReal(fieldcache, componentsCount)
+        assert result == RESULT_OK
+        del fieldNodesetMinimum
+        del fieldNodesetMaximum
+    return minimums, maximums
+
 def interpolateNodesCubicHermite(cache, coordinates, xi, normal_scale, \
```
hsorby: Functions like this utility should be in the `opencmiss.zinc.utils` package.

src/scaffoldmaker/utils/zinc_utils.py
```diff
@@ -354,33 +389,32 @@ def exnodeStringFromNodeValues(
...
+            for v in range(nodeValueLabelsCount):
+                coordinates.setNodeParameters(cache, -1, nodeValueLabels[v], 1, nodeValues[n][v])
+        # serialise to string
+        sir = region.createStreaminformationRegion()
+        srm = sir.createStreamresourceMemory()
+        sir.setResourceGroupName(srm, groupName)
+        region.write(sir)
+        result, exString = srm.getBuffer()
+        #print('\n',nodeValues)
+        #print('exnodeStringFromNodeValues', result, exString)
```
hsorby: Commented out debugging code should just be removed.

setup.py
```diff
@@ -40,7 +40,7 @@ def readfile(filename, split=False):
     url='',
     license='APACHE',
     packages=find_packages(exclude=['ez_setup',]),
```
hsorby: This line should be:
`packages=find_packages('src'),`

tests/test_heart.py
rchristie: Will fix.

src/scaffoldmaker/utils/zinc_utils.py
rchristie: I'm not sure what you mean here?

setup.py
rchristie: Will fix.

rchristie: You don't have it in mfxml2ex... what's the best usage here?
https://github.com/hsorby/mbfxml2ex/blob/master/setup.py uses:
`py_modules=["mbfxml2ex"],`
`package_dir={'': 'src'},`
Scaffoldmaker would have:
`packages=find_packages('src'),`
`package_dir={'': 'src'},`

hsorby: No mbfxml2ex doesn't have any packages so it really doesn't need to use the `find_packages` helper function.

src/scaffoldmaker/utils/zinc_utils.py
hsorby: These functions are all pretty much the same there is a lot of code that could be reused.

src/scaffoldmaker/utils/zinc_utils.py
rchristie: Will fix.

PR #68, opened 2020-02-13
Implement phase

mlin865:


rchristie:
Some of the test answers have changed – is this because the addition of phase has slightly changed the zero phase answer?

mlin865:
Yes, this is due to the different method I applied to calculate node locations now and given that the results are still quite close to the original values, I am quite happy with the changes.

rchristie:
You'll need to catch up again to get other commits.

PR #69, opened 2020-04-16
2d annotation

zekh167:


src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -825,6 +825,36 @@ def generateBaseMesh(region, options):
...
                 elementIdentifier += 1
 
+        # create 2d mesh groups
+        fm.defineAllFaces()
+        bodyGroup.addSubelements()
+        neckGroup.addSubelements()
+        mesh2d = fm.findMeshByDimension(2)
+        is_exterior = fm.createFieldIsExterior()
+        is_exterior_face_xi3_1 = fm.createFieldAnd(is_exterior, fm.createFieldIsOnFace(Element.FACE_TYPE_XI3_1))
+        is_interior = fm.createFieldIsExterior()
```
rchristie, line 835: Remove is_interior and just use is_exterior. Exterior here means on a surface or edge of an element without an adjacent element on any side, so it applies on the inside and outside of a walled body such as the bladder.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -825,6 +825,36 @@ def generateBaseMesh(region, options):
...
+        is_neck = neckGroup.getFieldElementGroup(mesh2d)
+        is_neck_outer_layer = fm.createFieldAnd(is_neck, is_exterior_face_xi3_1)
+        is_neck_lumen = fm.createFieldAnd(is_neck, is_interior_face_xi3_0)
+
+        serosaOfBody = AnnotationGroup(region, "Serosa of urinary bladder", FMANumber='unknown', lyphID='unknown')
+        serosaOfBody.getMeshGroup(mesh2d).addElementsConditional(is_body_serosa)
+        lumenOfBody = AnnotationGroup(region, "Lumen of body of urinary bladder", FMANumber='unknown', lyphID='unknown')
+        lumenOfBody.getMeshGroup(mesh2d).addElementsConditional(is_body_lumen)
+
+        outerLayerOfNeck = AnnotationGroup(region, "Neck of urinary bladder", FMANumber='unknown', lyphID='unknown')
```
rchristie, line 851: Is it really correct to call it the "Neck of urinary bladder" as it is the outer layer? You already have an annotation group called "neck of bladder".

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -825,6 +825,36 @@ def generateBaseMesh(region, options):
                 bodyMeshGroup.addElement(element)
                 elementIdentifier += 1
 
+        # create 2d mesh groups
+        fm.defineAllFaces()
+        bodyGroup.addSubelements()
```
rchristie, line 830: Change these lines to the following to pick up all annotation groups you have now and any you add later:
```
for annotationGroup in annotationGroups:
    annotationGroup.addSubelements()
```

PR #71, opened 2020-05-03
Transformations in ScaffoldPackage

rchristie:


src/scaffoldmaker/scaffoldpackage.py
```diff
@@ -75,12 +92,116 @@ def getScaffoldSettings(self):
...
+        assert len(translation) == 3
+        if self._translation != translation:
+            self._translation = translation
+            return True
+        return False
+
+    def getTransformationMatrix(self):
+        '''
+        :return: 4x4 row-major transformation matrix with first index down rows, second across columns,
+        suitable for multiplication p' = Mp where p = [ x, y, Z, h ], or None if identity.
```
mahyar-osn, line 140: Capital Z to small z.

src/scaffoldmaker/scaffoldpackage.py
```diff
@@ -75,12 +92,116 @@ def getScaffoldSettings(self):
...
+                [ rotationMatrix[1][0]*self._scale[0], rotationMatrix[1][1]*self._scale[1], rotationMatrix[1][2]*self._scale[2], self._translation[1] ],
+                [ rotationMatrix[2][0]*self._scale[0], rotationMatrix[2][1]*self._scale[1], rotationMatrix[2][2]*self._scale[2], self._translation[2] ],
+                [ 0.0, 0.0, 0.0, 1.0 ] ]
+        if not (all((v == 1.0) for v in self._scale) and all((v == 0.0) for v in self._translation)):
+            return [
+                [ self._scale[0], 0.0, 0.0, self._translation[0] ],
+                [ 0.0, self._scale[1], 0.0, self._translation[1] ],
+                [ 0.0, 0.0, self._scale[2], self._translation[2] ],
+                [ 0.0, 0.0, 0.0, 1.0 ] ]
+        return None
```
mahyar-osn, line 156: This may not be relevant, but was wondering if scaling would still work as before if rotation and translation are both 0 but the scale is changed?

rchristie: It should be handled by the second case, just with zeros in the appropriate column.

PR #72, opened 2020-05-04
Bladder annotation

zekh167:


src/scaffoldmaker/annotation/bladder_terms.py
```diff
@@ -4,9 +4,13 @@
...
-    ( "urinary bladder", "FMA:15900", "UBERON:0001255" ),
-    ( "neck of urinary bladder", "FMA:15912", "UBERON:0001258"),
-    ( "body of urinary bladder", None)  # needs to be replaced with an actual term e.g. urinary bladder (which includes neck)
+    ("urinary bladder", "FMA:15900", "UBERON:0001255"),
+    ("neck of urinary bladder", "FMA:15912", "UBERON:0001258"),
+    ("Dome of the Bladder", None),
+    ("Serosa of body of urinary bladder", None),
+    ("Lumen of body of urinary bladder", None),
+    ("Serosa of neck of urinary bladder", None),
+    ("Lumen of neck of urinary bladder", None)
```
rchristie, line 13: Need to put in the string with the term identifier, in this case "ILX:XXXXXXXX" since these are SPARC terms.
https://scicrunch.org/scicrunch/interlex/view/ilx_0738433
Can search for these.
https://scicrunch.org/scicrunch/interlex/search?q=serosa+bladder
Also, please use exact capitalisation of the text from interlex, e.g. Serosa and Lumen are not capitalised I think.
If you change them here, ensure they are changed everywhere.
"Dome of the Bladder" is an unusual choice, but it's in there; I would have expected "dome of urinary bladder".

zekh167:
Hi Richard,

all done. But can not do anything regarding "Dome of the Bladder". This
term was in the Scicrunch list before and I didn't request for that.

rchristie:
I was just saying it’s a pity these inconsistent names have crept in, nothing we can do about it now.
I’ll have a look at the PR again now.
Richard

zekh167:
Yeah, you're right. Even its capitalization is different from other terms!

Thanks Richard

PR #76, opened 2020-05-26
Cecum

mlin865:


src/scaffoldmaker/meshtypes/meshtype_3d_cecum1.py
```diff
@@ -0,0 +1,1059 @@
...
+            elementsCountRadial = 2, tracksurface = trackSurfaceOstium, startProportion = startProportion,
+            endProportion = endProportion)
+
+        # Delete elements under annulus mesh
+        deleteElementsAndNodesUnderAnnulusMesh(fm, nodes, mesh, deleteElementIdentifier, deleteNodeIdentifier)
+
+        return annotationGroups
+
+    @classmethod
+    def generateMesh(cls, region, options):
```
rchristie, line 677: Remove so Scaffold_base.generateMesh() is used and implement @classmethod def refineMesh(cls, meshRefinement, options) which will be similar to this function but without any call to generateBaseMesh.
I suggest doing this for all your scaffolds once you see how it works.

src/scaffoldmaker/meshtypes/meshtype_3d_cecum1.py
```diff
@@ -0,0 +1,1059 @@
...
+        #     xFlat, d1Flat, d2Flat, xTexture, d1Texture, d2Texture = tubemesh.createFlatAndTextureCoordinates(
+        #         xiList, relaxedLengthList, length, wallThickness, elementsCountAround,
+        #         elementsCountAlong, elementsCountThroughWall, transitElementList)
+        #
+        #     # Create nodes and elements
+        #     nextNodeIdentifier, nextElementIdentifier, annotationGroups = tubemesh.createNodesAndElements(
+        #         region, xList, d1List, d2List, d3List, xFlat, d1Flat, d2Flat, xTexture, d1Texture, d2Texture,
+        #         elementsCountAround, elementsCountAlong, elementsCountThroughWall,
+        #         annotationGroups, annotationArray, firstNodeIdentifier, firstElementIdentifier,
+        #         useCubicHermiteThroughWall, useCrossDerivatives)
```
rchristie, line 471: Remove commented out code, especially if can copy from source again.

src/scaffoldmaker/meshtypes/meshtype_3d_cecum1.py
```diff
@@ -0,0 +1,1059 @@
...
+            'Start tenia coli width derivative',
+            'End tenia coli width',
+            'End tenia coli width derivative',
+            'Ileocecal junction angular position degrees',
+            'Ileocecal junction position along factor',
+            'Tenia coli thickness',
+            'Wall thickness']:
+            if options[key] < 0.0:
+                options[key] = 0.0
+
```
rchristie, line 243: Implement a function like that in heartatria to transfer settings (ostium wall thickness etc.) to the ostium:
```
    @classmethod
    def updateSubScaffoldOptions(cls, options):
```
Call this function in each of getDefaultOptions, checkOptions, generateBaseMesh as in heartatria1.
This ensures the dependent ostium options are already up-to-date.
I'm not sure if the number around can be set, but see if the logic can be moved here.

src/scaffoldmaker/meshtypes/meshtype_3d_cecum1.py
```diff
@@ -0,0 +1,1059 @@
...
+            'Segment length mid derivative factor': 4.0,
+            'Number of tenia coli': 3,
+            'Start tenia coli width': 5.0,
+            'Start tenia coli width derivative': 0.0,
+            'End tenia coli width': 5.0,
+            'End tenia coli width derivative': 0.0,
+            'Tenia coli thickness': 0.5,
+            'Wall thickness': 2.0,
+            'Ileocecal junction': copy.deepcopy(ostiumOption),
+            'Ileocecal junction angular position degrees': 60,
```
rchristie, line 124: What happens if it's put over the tenia?

mlin865: Since it does not happen in rea life, I have an assert statement which prevents that from happening:

assert angleAroundInSector > angleToTCEdge + angleOstium * 0.5 and \
               angleAroundInSector < (2*math.pi/tcCount) - angleToTCEdge - angleOstium*0.5,\
               'Ileocecal junction cannot sit on tenia coli'

rchristie: Make sure all asserts like this (I see there are others) are avoided in proper use by checking and limiting the value in checkOptions; asserts are exceptions which shouldn't happen by something a user enters. It's fine to have them in generateBaseMesh to catch callers who misuse the library by setting them wrong not calling checkOptions to limit them.

src/scaffoldmaker/meshtypes/meshtype_3d_cecum1.py
mlin865: I made the changes for ostium wall thickness, but the number of elements around ostium is only determined much later after we have created the cecum segments and identified the tracksurface where ostium sits, so I don't think it makes sense to move it to this function.  Also, the number of elements around is derived and updated right before I make the ostium so it will ensure that the option is up-to-date:
ostiumSettings['Number of elements around ostium'] = len(innerEndPoints_Id)

src/scaffoldmaker/meshtypes/meshtype_3d_cecum1.py
mlin865: I couldn't limit the angular position of the ileocecal junction in checkOptions as I need to make the cecum segments first before I can use it to calculate the arclength of tenia coli, that's why I put it as an assert when the junction is found sitting over the tenia coli.  

It's the same for the other assert which identifies insufficient elements around the ostium to create a 2-radial elements annulus. I would only know that there is insufficient elements to make that annulus after I have created a tracksurface and know where to place the ostium.

mlin865:
All requested changes made/addressed and added a rudimentary test for cecum.  Just to let you know, there is an error in unit test for bladder which should be fixed.

PR #80, opened 2020-07-20
Multiple annotation groups

mlin865:


rchristie:
Can you walk me through this. I'm unsure why groups are being looked up by name when object can be used more efficiently? Also wondering about the ramifications of having a separate list for along, around and through the wall. What if there were different groups through the wall for some elements around?
Even if all is well there are some pythonic tricks that will make the code cleaner e.g.
```if name in names:```
(where name is a str and names is a list of str).

File: src/scaffoldmaker/meshtypes/meshtype_3d_colonsegment1.py [hunk 1]
```diff
@@ -1835,11 +1857,18 @@ def createNodesAndElementsTeniaColi(region,
...
+                            meshGroup = annotationGroup.getMeshGroup(mesh)
+                            meshGroup.addElement(element)
+                    for annotationAlong in annotationArrayAlong[0]:
+                        if annotationAlong == annotationGroup._name:
+                            meshGroup = annotationGroup.getMeshGroup(mesh)
+                            meshGroup.addElement(element)
+                    for annotationThroughWall in annotationArrayThroughWall[e3]:
+                        if annotationThroughWall == annotationGroup._name:
+                            meshGroup = annotationGroup.getMeshGroup(mesh)
+                            meshGroup.addElement(element)
```
rchristie, line 1871: You shouldn't directly use members prefixed with an underscore e.g. annotationGroup._name as this is the convention for 'private' (but nothing is truly private in python).
Would at a minimum use a method to get name and can replace lines 1860-1861 with (I think this works, but actually suggest going further below):
```
                        if annotationGroup.getName() in annotationArrayAround[e1]:
```
(Incidentally is it correct that the index is [0] in line 1864; maybe add a dummy e2 set to 0?)
If annotationArrayAround[e1] was a list of AnnotationGroup can replace lines 1860 to 1863 with:
```
                        if annotationGroup in annotationGroupsAround[e1]:
                            meshGroup = annotationGroup.getMeshGroup(mesh)
                            meshGroup.addElement(element)
```
Another alternative is to just put the lists from each array into a common lists:
```
annotationGroups = annotationGroupsAround[e1] + annotationGroupsAlong[e2] + annotationGroupsThroughWall[e3]
for annotationGroup in annotationGroups:
    meshGroup = annotationGroup.getMeshGroup(mesh)
    meshGroup.addElement(element)
```
If the same group appears twice the second call to addElement returns a benign error. See:
http://opencmiss.org/documentation/apidoc/zinc/latest/classOpenCMISS_1_1Zinc_1_1MeshGroup.html#ab0fa0e1b8ba89a3543c75c4c4c70f64d

mlin865:
Thanks for your suggestions, Richard. The code looks alot more elegant now with the revision. Please review the changes.

src/scaffoldmaker/meshtypes/meshtype_3d_colon1.py
```diff
@@ -441,47 +441,41 @@ def generateBaseMesh(cls, region, options):
...
-        annotationGroupsAlong = [colonGroup]
 
         if tcCount == 1:
             proximalGroup = AnnotationGroup(region, get_colon_term("proximal colon"))
             transverseGroup = AnnotationGroup(region, get_colon_term("transverse colon"))
             distalGroup = AnnotationGroup(region, get_colon_term("distal colon"))
-            annotationGroupsAlong += [proximalGroup, transverseGroup, distalGroup]
-            annotationGroupNamesAlong = [['colon', 'proximal colon'],
-                                         ['colon', 'transverse colon'],
-                                         ['colon', 'distal colon']]
```
rchristie: Are the names along/around/through wall still needed? Remove if not.

mlin865: The names are no longer needed, which is why they have been deleted.

PR #85, opened 2020-08-18
Bladder urethra

zekh167:


src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -0,0 +1,923 @@
...
+            'Central path LUT': copy.deepcopy(centralPathOption_LUT),
+            'Central path bladder': copy.deepcopy(centralPathOption_bladder),
+            'Number of elements along': 20,
+            'Number of elements around': 12,
+            'Number of elements through wall': 1,
+            'Major diameter': 30.0,
+            'Minor diameter': 25.0,
+            'Neck diameter 1': 5.0,
+            'Neck diameter 2': 3.0,
+            'Bladder wall thickness': 0.5,
```
rchristie, line 192: This is where not having sub-scaffolds for bladder and urethra is becoming problematic since you have to be clear about which part the parameter belongs to. Perhaps make it a rule that since this is 'Bladder with Urethra', bladder-only parameters don't need to be qualified, but you need to put 'Urethra' in front of urethra-specific parameters.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -0,0 +1,923 @@
...
+            'Neck angle': 45,
+            'Include urethra': True,
```
rchristie, line 194: If we have annotation groups on the central path, then making the urethra group empty could indicate remove the urethra?

rchristie (changes requested):
First the straightforward fixes:
There are some minor fixes in names and individual lines of code.
The separate utility functions need polishing (I'm quite careful about these).
While I think in the medium term I'd want the bladder in a separate scaffold for re-use, I think it's more important right now to update this code to use a common central path for the bladder and urethra with annotation groups marking where one stops and the other begins. We can generalise 1d path utility function extractPathParametersFromRegion to take an optional group and otherwise use it as before.

src/scaffoldmaker/utils/annulusmesh.py
```diff
@@ -462,3 +463,43 @@ def createAnnulusMesh3d(nodes, mesh, nextNodeIdentifier, nextElementIdentifier,
     fm.endChange()
 
     return nodeIdentifier, elementIdentifier
+
+def deleteElementsAndNodesUnderAnnulusMesh(fm, nodes, mesh, deleteElementIdentifier, deleteNodeIdentifier):
+    """
+    Deletes elements and nodes on tracked surface under annulus mesh using element and node identifiers.
+    :param deleteElementIdentifier: Element identifiers for elements to be deleted.
+    :param deleteNodeIdentifier: Node identifiers for nodes to be deleted.
+    """
```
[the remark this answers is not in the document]
zekh167: I talked to Mabelle and she was happy with the new changes.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -0,0 +1,923 @@
...
+            'Urethra diameter 1': 1.5,
+            'Urethra diameter 2': 1.0,
+            'Urethra wall thickness': 0.5,
+            'Length factor': 0.5,
```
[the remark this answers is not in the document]
zekh167: I changed it to "Urethra length proportion to central path". This parameter is used to identify the urethra length and also the bladder length along the central path.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -0,0 +1,879 @@
...
+
+        fm = region.getFieldmodule()
+        fm.beginChange()
+        nodes = fm.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+        mesh = fm.findMeshByDimension(3)
+
+        # Central path
+        tmpRegion = region.createRegion()
+        centralPath.generate(tmpRegion)
+        cx, cd1, cd2, cd12 = extractPathParametersFromRegion(tmpRegion)
```
rchristie, line 329: This will be called twice to extract parameters for bladder and urethra separately, passing the group name. See comment earlier.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -0,0 +1,879 @@
...
+            sx, sd1, se, sxi, ssf = interp.sampleCubicHermiteCurves(cx, cd1, elementsCountAlong)
+            sd2, sd12 = interp.interpolateSampleCubicHermite(cd2, cd12, se, sxi, ssf)
+        else:
+            m = int((1 - urethraproportion) * len(cx))
+            cx = cx[:m+1]
+            cd1 = cd1[:m+1]
+            cd2 = cd2[:m+1]
+            cd12 = cd12[:m+1]
+            sx, sd1, se, sxi, ssf = interp.sampleCubicHermiteCurves(cx, cd1, elementsCountAlongBladder)
+            sd2, sd12 = interp.interpolateSampleCubicHermite(cd2, cd12, se, sxi, ssf)
```
rchristie, line 364: Not needed with annotated central path.

zekh167: If I delete this part, how I can run the following part?

        # Project reference point for warping onto central path
        sxRefList, sd1RefList, sd2ProjectedListRef, zRefList = \
             tubemesh.getPlaneProjectionOnCentralPath(innerNodes_x, elementsCountAround, elementsCountAlong,
                                                      length, sx, sd1, sd2, sd12)

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -0,0 +1,879 @@
...
+            'Minor diameter',
+            'Neck diameter 1',
+            'Neck diameter 2',
+            'Wall thickness',
+            'Neck angle degrees',
+            'Include urethra',
+            'Urethra diameter 1',
+            'Urethra diameter 2',
+            'Urethra wall thickness',
+            'Urethra length proportion to central path',
```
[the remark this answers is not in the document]
zekh167: After deleting the "Urethra length proportion to central path", Should I delete the "Number of elements Along" from options and instead add "Number of elements Along Bladder" and "Number of elements Along Urethra", which are independent? 
Then, the segmentLength will be different in bladder and urethra parts, right?

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
zekh167: Should I change the following part, too? Because the segmentLength will be different after deleting the "urethraLengthPortion". 

        # Find arcLength
        length = 0.0
        elementsCountIn = len(cx) - 1
        sd1 = interp.smoothCubicHermiteDerivativesLine(cx, cd1, fixAllDirections=True,
                                                       magnitudeScalingMode=interp.DerivativeScalingMode.HARMONIC_MEAN)
        for e in range(elementsCountIn):
            arcLength = interp.getCubicHermiteArcLength(cx[e], sd1[e], cx[e + 1], sd1[e + 1])
            length += arcLength
        segmentLength = length / elementsCountAlong

src/scaffoldmaker/utils/geometry.py
```diff
@@ -242,3 +242,31 @@ def getSurfaceProjectionAxes(ax, ad1, ad2, ad3, angle1radians, angle2radians, le
     bd1 = vector.crossproduct3(ad2, bd3)
     bd2 = vector.crossproduct3(bd3, bd1)
     return bx, bd1, bd2, bd3
+
+def createEllipsePoints(cx, radian, axis1, axis2, elementsCountAround, startRadians = 0.0):
```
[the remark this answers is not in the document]
rchristie: I still want createEllipsePoints changed as described above.

src/scaffoldmaker/utils/annulusmesh.py
```diff
@@ -6,8 +6,9 @@
 import copy
 import math
 from opencmiss.utils.zinc.field import findOrCreateFieldCoordinates
+from opencmiss.utils.zinc.general import ChangeManager
 from opencmiss.zinc.element import Element
-from opencmiss.zinc.field import Field
+from opencmiss.zinc.field import Field, FieldGroup
 from opencmiss.zinc.node import Node
```
rchristie, line 12: Remove the remnants to your changes to annulus mesh since you've moved element/node deletion code elsewhere.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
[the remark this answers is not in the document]
rchristie: Yes, you'll have independent numbers of elements along bladder and urethra, but you need to make the element size transition properly between them. Either the urethra fits to the bladder or vice versa.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
rchristie: Shouldn't need to smooth the central path. (Separately we need to find a way for the user to smooth the 1D path as needed; on the to-do list.)
You should have a separate cx, cd1 for bladder and urethra. Can have a fixed segment length for one or the other, but not both:
If it is the bladder, the first element of the urethra needs to transition from the derivatives supplied by the end of the bladder. If it is the urethra, the last element of the bladder needs to transition to fit the first urethra derivatives.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
rchristie: You can still do this on each respective sets of cx, cd1 for the bladder and urethra.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
rchristie: I'm happy with 'Include urethra', easy to use.

rchristie (approved):
Good.
At some future point we may want to use some of the new smoothing/normal functions of 1d path to improve default central paths.

PR #87, opened 2020-08-20
Annotation fix

rchristie:


mahyar-osn (approved):
Looks good. Thanks for the quick fix.

Just something probably not important and urgent now with `fieldmodule` `begin `and `end`. Would it be better to change all the `meshtypes ` so that they all use the new `with changeManager` statement?

rchristie:
At some point I'd like to push through full use of ChangeManager, but not now. I may remove it from scaffolds altogether and make it the client's job.
I just noticed ScaffoldPackage.generate doesn't do ChangeManager... it probably should.

PR #89, opened 2020-09-02
Stellate

s-fong:


src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -0,0 +1,543 @@
...
+    """
+
+    elementsCount1, elementsCount2, elementsCount3 = elementsCount
+    [elen, ewid, zheight] = elementLengths
+
+    xnodes_ds1 = []
+    xnodes_ds2 = []
+    dx_ds1 = [elen, 0.0, 0.0]
+    dx_ds2 = [0.0, ewid, 0.0]
+    wheelDvMult = [0.5, 0.75]
```
rchristie, line 419: These factors should be calculated from smoothing around the gap between arms. Can fix it up after the main loop and use smoothCubicHermiteDerivativesLine to fix d1 at the dips.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -0,0 +1,543 @@
...
+    if na == 0:
+        rmVertexNodes = []
+    elif na != numArm - 1:
+        rmVertexNodes = nCentre + [0, nodes_per_layer]
+    else:
+        rmVertexNodes = nCentre + [0, nodes_per_layer,
+                          2* (elementsCount1+1) - 1, 2* (elementsCount1+1) - 1 + nodes_per_layer ]
+    dcent = [elen * math.cos(armAngleConst / 2), elen * math.sin(armAngleConst / 2), 0.0]
+    dvertex = [round(elen * dipMultiplier * math.cos(thAr), 12),
+                round(elen * dipMultiplier * math.sin(thAr), 12)]
```
rchristie, line 436: This should be mean of element length and width by default.
```
    dipLength = 0.5*(elen + ewid)*dipMultiplier
    dvertex = [round(dipLength * math.cos(thAr), 12),
               round(dipLength * math.sin(thAr), 12)]
    dipMag = 2*dipLength - elen
```
Magnitude of d2 on outside should later be set to dipMag.
Why are these rounded? Avoid?

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -0,0 +1,543 @@
...
+                        ds2 = dcent
+                        ds1 = [dcent[0], -dcent[1], dcent[2]]
+                    elif n1 == 0 and (n2 == 0 or n2 == elementsCount2):
+                        [p, q] = [x1/dipMultiplier, x2/dipMultiplier] # unit vector
+                        ds2mag = round(2*magnitude([p,q]),6) - magnitude(dcent)
+                        ds2 = [ds2mag*p, ds2mag*q, 0]
+                        ds1 = rotateAboutZAxis(ds2, -math.pi / 2)
+                        for i in range(2):
+                            ds1[i] *= wheelDvMult[0]
+                        ds1 = [d * elen for d in ds1]
```
rchristie, line 471: d2 code:
```
ds2 = [dcent[0], -dcent[1], dcent[2]] if (n2 == 0) else dcent
ds2 = setMagnitude(ds2, dipMag)  # must import function from utils.vector
```

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -0,0 +1,543 @@
...
+    @staticmethod
+    def getDefaultOptions(parameterSetName='Default'):
+        return {
+            'Number of arms' : 3,
+            'Number of elements in long arm' : 4,
+            'Number of elements in short arms' : 2,
+            'Element length along arm' : 1.0,
+            'Element width across arm' : 0.5,
+            'Element thickness' : 0.5,
+            'Extent of indentation around centre' : 0.5,
```
rchristie, line 35: This is not easy to understand. Why not have a scale factor on the dip between arms, with 1.0 the default which makes the dip 0.5*(elementLength + elementWidth) away -- see code below which does this.

rchristie (changes requested):
Some fixes needed.
I'm also interested in the node at the end of the arm being a reduced distance from the next node so that final element length is 0.5*(elementLength + elementWidth) and d1 is magnitude elementWidth i.e. matching d2 around the whole outside (which is elegant). Maybe later. That would mean d2 wouldn't have to be so large here.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
s-fong: I am unsure which lines should be replaced by your d2 code: it doesn't seem to work. And did you mean dipMag = ds2mag?

rchristie: I set dipMag in a comment earlier...
The above two lines of mine entirely calculate d2, but you defined d1 from d2 (or at least rotate it 90 degrees). As mentioned in a further comment, it's fine to get the direction of d1 at the dip here, but use the smooth function to get its magnitude later - don't use arbitrary scalings.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -0,0 +1,543 @@
...
+            'Refine number of elements along arm',
+            'Refine number of elements across arm',
+            'Refine number of elements through thickness'
+        ]
+
+    @staticmethod
+    def checkOptions(options):
+        for key in [
+            'Number of elements in long arm',
+            'Number of elements in short arms',
```
rchristie, line 62: Shouldn't these be limited to >=2?

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
s-fong: How do I go about replacing this code?

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -69,16 +62,19 @@ def checkOptions(options):
...
 
-        if options['Number of elements in short arms'] < 2:
-            options['Number of elements in short arms'] = 2
-
-        # if options['Number of arms'] < 3 or options['Number of arms'] > 4:
-        #     options['Number of arms'] = 3
+        key = 'Number of elements in all arms (e.g. 4,2,2)'
+        if len(options[key]) < 3:
+            x = [2]*(3 - len(options[key]))
+            options[key].append(x)
```
rchristie, line 72: This doesn't actually work; line 2 (x = ...) makes a list, line 3 appends it to the list i.e. adding a last element that is itself a list. You could change the last line to options[key] += x
However, I would do this differently to make it ready for variable numbers of arms, and to avoid the issue where if you have 2 missing arms, the code may make the last 2 values references to the same value 2:
```
armCountsKey = 'Numbers of elements along arms'
armCount = 3  # to allow for it being a parameter later
countsLen = len(options[armCountsKey]
if countsLen < armCount:
    options[armCountsKey] += [ 2 for i in range(armCount - countsLen) ]
elif countsLen > armCount:
    options[armCountsKey] = options[armCountsKey][:armCount]
'''

rchristie (changes requested):
Some fairly minor fixes now - see inline notes (which I put on the 2nd last commit if not easy to find).
Can you also please remove fieldmodule beginChange() and endChange() calls from generateBaseMesh(). I no longer want these forms used and the replacement ChangeManager(fieldmodule) should already be done by the code calling this.

rchristie (approved):
Great.

PR #91, opened 2020-09-10
Add scaffold interactive functions

rchristie:
Allow 1D path parameters to be extracted by group.

### DIFF
--- a/src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
+++ b/src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
@@ -5,10 +5,12 @@ Generates a 1-D path mesh.
 from __future__ import division
 import math
 from opencmiss.utils.zinc.field import findOrCreateFieldCoordinates
+from opencmiss.utils.zinc.general import ChangeManager
 from opencmiss.zinc.element import Element, Elementbasis
 from opencmiss.zinc.field import Field
 from opencmiss.zinc.node import Node
 from scaffoldmaker.meshtypes.scaffold_base import Scaffold_base
+from scaffoldmaker.utils.interpolation import DerivativeScalingMode, smoothCubicHermiteDerivativesLine
 
 class MeshType_1d_path1(Scaffold_base):
     '''
@@ -54,16 +56,15 @@ class MeshType_1d_path1(Scaffold_base):
         length = options['Length']
         elementsCount = options['Number of elements']
 
-        fm = region.getFieldmodule()
-        fm.beginChange()
-        coordinates = findOrCreateFieldCoordinates(fm, components_count=coordinateDimensions)
-        cache = fm.createFieldcache()
+        fieldmodule = region.getFieldmodule()
+        coordinates = findOrCreateFieldCoordinates(fieldmodule, components_count=coordinateDimensions)
+        cache = fieldmodule.createFieldcache()
 
         #################
         # Create nodes
         #################
 
-        nodes = fm.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+        nodes = fieldmodule.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
         nodetemplate = nodes.createNodetemplate()
         nodetemplate.defineField(coordinates)
         nodetemplate.setValueNumberOfVersions(coordinates, -1, Node.VALUE_LABEL_VALUE, 1)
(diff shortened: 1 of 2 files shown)


src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
```diff
@@ -103,28 +104,50 @@ def generateBaseMesh(cls, region, options):
...
     cd12 = []
-    nodes = fm.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+    nodes = fieldmodule.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+    if groupName:
+        group = fieldmodule.findFieldByName(groupName).castGroup()
+        nodeGroup = group.getFieldNodeGroup(nodes)
+        if nodeGroup.isValid():
+            nodes = nodeGroup.getNodesetGroup()
+        else:
+            print('extractPathParametersFromRegion: missing group "' + groupName + '"')
```
elias-soltani, line 150: Shouldn't be under ```if groupName:``` ?

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
```diff
@@ -103,28 +104,50 @@ def generateBaseMesh(cls, region, options):
...
+    cache = fieldmodule.createFieldcache()
     cx = []
     cd1 = []
     cd2 = []
     cd12 = []
-    nodes = fm.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+    nodes = fieldmodule.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+    if groupName:
+        group = fieldmodule.findFieldByName(groupName).castGroup()
+        nodeGroup = group.getFieldNodeGroup(nodes)
```
elias-soltani, line 146: gyt

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
rchristie: Note sure what you mean. I just wanted a warning if the group was not found (meaning anything from not existing to having no nodes in it). If not group name is specified, it uses all nodes; if not found it also uses all nodes.

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
```diff
@@ -103,28 +104,50 @@ def generateBaseMesh(cls, region, options):
...
         return []
 
+    @classmethod
+    def smoothPath(cls, region, options, mode : DerivativeScalingMode):
+        x, d1 = extractPathParametersFromRegion(region)[0:2]
+        d1 = smoothCubicHermiteDerivativesLine(x, d1, magnitudeScalingMode=mode)
+        setPathParameters(region, [ Node.VALUE_LABEL_VALUE, Node.VALUE_LABEL_D_DS1 ], [ x, d1 ])
+
+    @classmethod
+    def getInteractiveFunctions(cls):
```
elias-soltani, line 116: Nice!

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
```diff
@@ -103,28 +104,50 @@ def generateBaseMesh(cls, region, options):
...
     componentsCount = coordinates.getNumberOfComponents()
     assert componentsCount in [ 1, 2, 3 ], 'extractPathParametersFromRegion.  Invalid coordinates number of components'
-    cache = fm.createFieldcache()
+    cache = fieldmodule.createFieldcache()
     cx = []
     cd1 = []
     cd2 = []
     cd12 = []
-    nodes = fm.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
+    nodes = fieldmodule.findNodesetByFieldDomainType(Field.DOMAIN_TYPE_NODES)
```
elias-soltani, line 143: I think it should be as part of the first if statement

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
elias-soltani: I think it should be a part of the first if statement.

elias-soltani: I might be wrong

elias-soltani: This is what I think it should be

```
    if groupName:
        group = fieldmodule.findFieldByName(groupName).castGroup()
        nodeGroup = group.getFieldNodeGroup(nodes)
        if nodeGroup.isValid():
            nodes = nodeGroup.getNodesetGroup()
    else:
         print('extractPathParametersFromRegion: missing group "' + groupName + '"')
```

rchristie: No, groupName is optional; if you don't supply it (or it is empty/doesn't exist) you get all nodes.
What may not be so obvious is that the NodesetGroup is a derived type, so the 3rd to last line is replacing all nodes with the subset of nodes, but the following code will work either way (because NodesetGroup implements all functions of the parent domain on its subset).

elias-soltani: Thanks for clearing that up.

elias-soltani (approved):

PR #92, opened 2020-09-17
Annotation

s-fong:


src/scaffoldmaker/annotation/stellate_terms.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+
+# convention: preferred name, preferred id, followed by any other ids and alternative names
+stellate_terms = [
+    ( "stellate arm 1", None),
+    ( "stellate arm 2", None),
+    ( "stellate arm 3", None),
+    ( "stellate face 2-3", None),
+    ( "stellate face 1-2", None),
+    ( "stellate face 3-1", None),
```
rchristie, line 12: It's OK to put these here I guess, but I'm doubtful about whether they'll get proper ontology term IDs.
Also, if you support N-arms on the stellate you'll have a harder time managing these; you could have just created the names in the code according to this pattern.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -85,7 +113,12 @@ def generateBaseMesh(cls, region, options):
         :param options: Dict containing options. See getDefaultOptions().
         :return: None
         """
+        isDefault = 'Default' in options['Base parameter set']
+        isMouse = 'Mouse' in options['Base parameter set'] and 'Long' not in options['Base parameter set']
+        isLongMouse = 'Long Mouse' in options['Base parameter set']
+
         armCount = 3
+        elementLengthCentral = options['Element length around central wheel']
```
rchristie, line 121: elementWidthCentral

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -367,8 +465,36 @@ def generateBaseMesh(cls, region, options):
...
+
+                        # add to meshGroup
+                        stellateMeshGroup.addElement(element)
+                        if isMouse:
+                            if na == 0:
+                                arm1MeshGroup.addElement(element)
+                            elif na == 1:
+                                arm2MeshGroup.addElement(element)
+                            elif na == 2:
+                                arm3MeshGroup.addElement(element)
```
rchristie, line 477: I think the arm and face groups could be defined for all stellates, not just mouse (the markers are just for mouse though).
Also, I'd prefer a more elegant implementation where you have an array of groups for arm index.
Try to avoid repeating code such as addElement; get the mesh groups by arm, then add the element with the same code for all cases.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -367,8 +465,36 @@ def generateBaseMesh(cls, region, options):
...
+                                if elementIdentifier == allMarkers[key]["elementID"]:
+                                    elementID = allMarkers[key]["elementID"]
+                                    xi = allMarkers[key]["xi"]
+                                    addMarker = {"name": key, "xi": allMarkers[key]["xi"]}
+
+                                    markerPoint = markerPoints.createNode(nodeIdentifier, markerTemplateInternal)
+                                    nodeIdentifier += 1
+                                    cache.setNode(markerPoint)
+                                    markerName.assignString(cache, addMarker["name"])
+                                    markerLocation.assignMeshLocation(cache, element, addMarker["xi"])
```
rchristie, line 493: Need to discuss if there's a better way to do this, especially if we locate markers by proportions over each arm. (I typically try to avoid having loops overall markers for all elements.)
I'm thinking the elements loop builds arrays of n*2 element identifies for each arm, then have a subsequent step outside the loop for making marker points on it.

rchristie (changes requested):
Some changes as noted. I'll review the face annotations later.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -108,27 +132,88 @@ def generateBaseMesh(cls, region, options):
...
+                            "Dorsal ansa subclavia" : {"elementID": elementIndex['DA']+2*(elementsCount1[0]+elementsCount1[1]), "xi": [xi1['DA'], 0.0, 0.5]},
+                            "Cervical spinal nerve 8" : {"elementID": elementIndex['C8']+2*(elementsCount1[0]+elementsCount1[1])+elementsCount1[2], "xi": [xi1['C8'], 1.0, 0.5]},
+                            "Thoracic spinal nerve 1" : {"elementID": elementIndex['T1'], "xi": [xi1['T1'], 0.0, 0.5]},
+                            "Thoracic spinal nerve 2" : {"elementID": elementIndex['T2'], "xi": [xi1['T2'], 0.0, 0.5]},
+                            "Thoracic spinal nerve 3" : {"elementID": elementIndex['T3'], "xi": [xi1['T3'], 0.0, 0.5]},
+                            "Thoracic sympathetic nerve trunk" : {"elementID": elementIndex['TST'], "xi": [xi1['TST'], 1.0, 0.5]},
+                           }
+
+        # arm group annotations for user
+        armNames, faceNames = getUnlabelledStellateArmNames(armCount)
```
rchristie, line 185: Use 'Term' instead of Name for all these as they are tuple (name, id). I expect 'Name' to only be a simple string.
Also, if you're not using one of the return values it's better to follow one of these idioms to ignore it:
```
armNames, _ = getUnlabelledStellateArmNames(armCount)
armNames = getUnlabelledStellateArmNames(armCount)[0]
```
The last one is preferred as it will work even if you add more return values.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -384,16 +490,65 @@ def refineMesh(cls, meshrefinement, options):
...
+            fm.createFieldAnd(is_arm1, is_exterior_face_xi2_1),
+            fm.createFieldAnd(is_arm2, is_exterior_face_xi2_0))
+        topStellateGroup = findOrCreateAnnotationGroupForTerm(annotationGroups, region, faceNames[0])
+        topStellateGroup.getMeshGroup(mesh2d).addElementsConditional(is_top_face)
+
+        is_bottom_face = fm.createFieldOr(
+            fm.createFieldAnd(is_arm1, is_exterior_face_xi2_0),
+            fm.createFieldAnd(is_arm3, is_exterior_face_xi2_1))
+        bottomStellateGroup = findOrCreateAnnotationGroupForTerm(annotationGroups, region, faceNames[-1])
+        bottomStellateGroup.getMeshGroup(mesh2d).addElementsConditional(is_bottom_face)
```
rchristie, line 535: You should be able to make a loop over arms and faces that would work with any number of arms AND be less code than this. The mesh2d, is_exterior etc. lines can go before loop, then (using the clever python trick of negative array indexes counting back from the end of the list):
```
armGroups = [getAnnotationGroupForTerm(annotationGroups, armTerm]) for armTerm in armTerms]
isArm = [armGroup.getFieldElementGroup(armGroup) for armGroup in armGroups]
for arm in armCount:
    is_face = fm.createFieldOr(
            fm.createFieldAnd(is_arm[arm - 1], is_exterior_face_xi2_1),
            fm.createFieldAnd(is_arm[arm], is_exterior_face_xi2_0))
    faceGroup = findOrCreateAnnotationGroupForTerm(annotationGroups, region, faceTerms[arm - 1])
    faceGroup.getMeshGroup(mesh2d).addElementsConditional(is_face)
```
DO check my logic.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -384,16 +490,65 @@ def refineMesh(cls, meshrefinement, options):
...
+
+def getUnlabelledStellateArmNames(armCount):
```
rchristie, line 537: getAutomaticArmFaceTerms?

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -384,16 +490,65 @@ def refineMesh(cls, meshrefinement, options):
...
+    armNames = [("stellate arm %d"%(d), None) for d in range(1,armCount+1)]
+    faceNames = [("stellate face %d-%d"%(d, d+1), None) for d in range(1,armCount+1)]
+    faceNames[-1] = ("stellate face %d-1"%(armCount), None)
```
rchristie, line 540: I suggest not using d for the indexes as it's confusing with the %d. Lines 2 and 3 can be merged using modulo operator %:
```
    faceNames = [("stellate face %d-%d"%(i, (i % armCount) + 1), None) for i in range(1, armCount + 1)]
```
(I haven't seen that way to make strings before... dinky!)
Nb. I prefer to see a space after a comma, and spaces around operators esp. + and - in expressions.

rchristie (changes requested):
In addition to comments in code.
I think the generic shape with 'default' parameters should have a generic annotation group 'stellate', not 'cervicothoracic ganglion'.
I'm thinking it would make sense to change parameter set 'Mouse 1' to be 'Mouse cervicothoracic ganglion 1', and make the group of that name for that case?

s-fong:
I am not changing the term 'cervicothoracic ganglion' to 'stellate'. 'Stellate' is unscientific and is more of a nickname, so I'm leaving the names as is.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -184,11 +184,13 @@ def generateBaseMesh(cls, region, options):
         # arm group annotations for user
         armTerms, _ = getAutomaticArmFaceTerms(armCount)
         armGroups = [AnnotationGroup(region, armTerm) for armTerm in armTerms]
-        stellateGroup = AnnotationGroup(region, get_stellate_term("cervicothoracic ganglion"))
-        annotationGroups = [stellateGroup] + armGroups
+        wholeScaffoldGroup = AnnotationGroup(region, ("stellate", None))
+        if isMouse:
+            wholeScaffoldGroup = AnnotationGroup(region, get_stellate_term("cervicothoracic ganglion"))
```
rchristie, line 189: I don't like needlessly making an object then remaking it (can be expensive in general) so prefer:
```
wholeScaffoldTerm = get_stellate_term("cervicothoracic ganglion") if isMouse else ("stellate", None)
wholeScaffoldGroup = AnnotationGroup(region, wholeScaffoldTerm)
```
Also, suggest rename these back to stellateTerm and stellateGroup, so consistent with stellateMeshGroup later. (Alternatively rename the latter to wholeScaffoldMeshGroup for consistency).

rchristie (changes requested):
A pretty minor pair of fixes now then it's in this morning.

rchristie (approved):

PR #93, opened 2020-09-21
Path1/sidederivatives

elias-soltani:
I would like to add some features to 1D path, please.

### DIFF
--- a/src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
+++ b/src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
@@ -132,7 +132,7 @@ class MeshType_1d_path1(Scaffold_base):
 
     @classmethod
     def smoothPath(cls, region, options, mode : DerivativeScalingMode):
-        x, d1 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_VALUE, Node.VALUE_LABEL_D_DS1])
+        x, d1 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_VALUE, Node.VALUE_LABEL_D_DS1])[0:2]
         d1 = smoothCubicHermiteDerivativesLine(x, d1, magnitudeScalingMode=mode)
         setPathParameters(region, [ Node.VALUE_LABEL_VALUE, Node.VALUE_LABEL_D_DS1 ], [ x, d1 ])
 
@@ -140,7 +140,7 @@ class MeshType_1d_path1(Scaffold_base):
     def makeD2Normal(cls, region, options):
         if not options['D2 derivatives']:
             return
-        d1, d2 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_D_DS1, Node.VALUE_LABEL_D_DS2])
+        d1, d2 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_D_DS1, Node.VALUE_LABEL_D_DS2])[0:2]
         for c in range(len(d1)):
             td2 = vector.vectorRejection(d2[c], d1[c])
             d2[c] = vector.setMagnitude(td2, vector.magnitude(d2[c]))
@@ -152,12 +152,12 @@ class MeshType_1d_path1(Scaffold_base):
             return
         if options['D2 derivatives']:
             d1, d2, d3 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_D_DS1, Node.VALUE_LABEL_D_DS2,
-                                                                  Node.VALUE_LABEL_D_DS3])
+                                                                  Node.VALUE_LABEL_D_DS3])[0:3]
             for c in range(len(d1)):
                 d3[c] = vector.setMagnitude(vector.crossproduct3(d1[c], d2[c]), vector.magnitude(d3[c]))
             setPathParameters(region, [Node.VALUE_LABEL_D_DS3], [d3])
         else:
-            d1, d3 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_D_DS1, Node.VALUE_LABEL_D_DS3])
+            d1, d3 = extractPathParametersFromRegion(region, [Node.VALUE_LABEL_D_DS1, Node.VALUE_LABEL_D_DS3])[0:2]
             for c in range(len(d1)):
                 td3 = vector.vectorRejection(d3[c], d1[c])
                 d3[c] = vector.setMagnitude(td3, vector.magnitude(d3[c]))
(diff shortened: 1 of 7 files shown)


src/scaffoldmaker/utils/vector.py
```diff
@@ -42,3 +42,50 @@ def addVectors(v1,v2,s1=1.0,s2=1.0):
...
+    """
+    Calculate s * v
+    :param s: Scalar.
+    :param v: Vector.
+    :return:
+    """
+    return [s * v[c] for c in range(len(v))]
+
+
+def isVectorsParallel(v1, v2):
```
rchristie, line 82: Suggest it's more robust to check whether cross product of vectors is less than some small tolerance of their mean magnitude.
Rename to parallelVectors().

src/scaffoldmaker/utils/vector.py
```diff
@@ -42,3 +42,50 @@ def addVectors(v1,v2,s1=1.0,s2=1.0):
     :return: Vector s1*v1+s2*v2
     '''
     return [(s1 * v1[c] + s2 * v2[c]) for c in range(len(v1))]
+
+
+def scalarProjectionOfV1OnV2(v1, v2):
```
rchristie, line 47: Rename to just scalarProjection. Same for following 2 functions, remove OfV1OnV2.

rchristie (changes requested):
Some small changes then it's in.

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
```diff
@@ -112,14 +136,44 @@ def smoothPath(cls, region, options, mode : DerivativeScalingMode):
...
+    def makeD3Normal(cls, region, options):
+        if not options['D3 derivative']:
+            return
+        if options['D2 derivative']:
+            x, d1, d2, d12, d3, d13 = extractPathParametersFromRegion(region)
+            for c in range(len(d1)):
+                d3[c] = vector.setMagnitude(vector.crossproduct3(d1[c], d2[c]), vector.magnitude(d3[c]))
+            setPathParameters(region, [Node.VALUE_LABEL_D_DS3], [d3])
+        else:
+            x, d1, d2, d12, d3, d13 = extractPathParametersFromRegion(region)
```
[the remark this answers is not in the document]
elias-soltani: Do you mean something like this? ```x, d1, _, _, d3, d13 = extractPathParametersFromRegion(region)```
Or do you mean the function should only return x,d1,d3 and d13?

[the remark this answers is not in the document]
rchristie: See comments on function; if you request, d1,d2,d3 that is what you get. Best not to extract values just to throw them away.

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
```diff
@@ -112,14 +136,44 @@ def smoothPath(cls, region, options, mode : DerivativeScalingMode):
...
         return [
             ("Smooth d1 arithmetic", lambda region, options: cls.smoothPath(region, options, DerivativeScalingMode.ARITHMETIC_MEAN)),
-            ("Smooth d1 harmonic", lambda region, options: cls.smoothPath(region, options, DerivativeScalingMode.HARMONIC_MEAN)) ]
+            ("Smooth d1 harmonic", lambda region, options: cls.smoothPath(region, options, DerivativeScalingMode.HARMONIC_MEAN)),
+            ("Make D2 normal", lambda region, options: cls.makeD2Normal(region, options)),
+            ("Make D3 normal", lambda region, options: cls.makeD3Normal(region, options))
+        ]
 
 
 def extractPathParametersFromRegion(region, groupName=None):
```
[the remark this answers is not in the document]
elias-soltani: I did not understand what you mean here. Do you mean like this?
```("Make D2 normal", lambda region, options: cls.makeD2Normal(region, nodeValueLabels, options))```
If yes, for D2 I need [Node.VALUE_LABEL_D_DS2] which right now is passed inside the function. Can you please tell me what do you mean?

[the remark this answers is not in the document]
Comment (rchristie) [reply src/scaffoldmaker/meshtypes/meshtype_1d_path1.py:179]: I meant only this function signature:
```
def extractPathParametersFromRegion(region, valueLabels, groupName=None)
```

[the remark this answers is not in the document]
Comment (elias-soltani) [reply src/scaffoldmaker/meshtypes/meshtype_1d_path1.py:179]: Thanks for elaboration. I will make the change.

**Comment by elias-soltani** [comment]
Thanks for reviewing the code.

File: src/scaffoldmaker/meshtypes/meshtype_1d_path1.py [hunk 5]
```diff
@@ -229,7 +232,22 @@ def extractPathParametersFromRegion(region, groupName=None):
...
+        if label == Node.VALUE_LABEL_D_DS1:
+            result += (cd1,)
+        if label == Node.VALUE_LABEL_D_DS2:
+            result += (cd2,)
+        if label == Node.VALUE_LABEL_D_DS3:
+            result += (cd3,)
+        if label == Node.VALUE_LABEL_D2_DS1DS2:
+            result += (cd12,)
+        if label == Node.VALUE_LABEL_D2_DS1DS3:
+            result += (cd13,)
```
rchristie: This isn't the solution I'm looking for; should instead build the result directly and loop over the number of valueLabels and call getNodeParameters only for those requested in the code earlier.

src/scaffoldmaker/utils/vector.py
```diff
@@ -79,13 +79,12 @@ def scalarProduct(s, v):
...
     :return: True if the vectors are parallel.
     """
     assert (len(v2) == len(v1)), 'Vectors lengths are not the same.'
-    s = magnitude(v2)/magnitude(v1)
-    for c in range(len(v2)):
-        if v2[c] != s * v1[c]:
-            return False
-    return True
+    TOL = 1.0e-6
+    if magnitude(crossproduct3(v1, v2)) < TOL * (magnitude(v1)+magnitude(v2))/2.0:
```
rchristie, line 88: Minor. No need to divide by 2 as you can halve TOL.

rchristie (changes requested):

src/scaffoldmaker/meshtypes/meshtype_1d_path1.py
elias-soltani: I modified the code so it only gets the values specified by valueLabels. What we return depends on the values specified by the valueLabels, I do not know if there is another way to do this. Can you please tell me what else can be done?

rchristie: Combine with the evaluation loop:
```
valueLabelsCount = len(valueLabels)
returnValues = ([] for i in range(valueLabelsCount)
while node.isValid():
        cache.setNode(node)
        for i in range(valueLabelsCount):
            result, values  = coordinates.getNodeParameters(cache, -1, valueLabels[i], 1, componentsCount)
            for c in range(componentsCount, 3):
                values.append(0.0)
            returnValues[i].append(values)
        node = nodeIter.next()
return returnValues
```

elias-soltani: Brilliant! Thanks.

rchristie (approved):

PR #94, opened 2020-09-29
Cylinder/central path

elias-soltani:
I would like to add the central path feature to cylinder scaffold, please.


src/scaffoldmaker/utils/centralpath.py
```diff
@@ -0,0 +1,77 @@
...
+    Sample elements along the path
+    :param cx: coordinates along the path.
+    :param cd1: d1 derivatives.
+    :param elementsCount: number of elements to be created.
+    :return: sx, sd1, se, sxi, ssf. coordinates and derivatives sampled along the path. also returns xi locations
+    and list of elements sampled.
+    """
+    sx, sd1, se, sxi, ssf = interp.sampleCubicHermiteCurves(cx, cd1, elementsCount)
+
+    return sx, sd1, se, sxi, ssf
```
rchristie, line 77: I wouldn't have made this file just for this.
getCentralPathNodes is useful but it is misnamed as it doesn't return nodes, and it belongs in the same file as extractPathParametersFromRegion (which you could move to this file and update callers). I wouldn't have print functionality in it as it slows all usage and can be done separately. If you're keeping this convenience function it should take the derivatives and group to request, so it's a 1 stop solution for colon, bladder too, otherwise it's only useful to your code.
calculateTotalLength is useful but belongs in interpolation.py; I'd call it 'getCubicHermiteCurvesLength'. Do you have a long term need to print arc lengths? If not remove.
The other 2 functions do not add anything to the interpolation functions they call.

rchristie (changes requested):
It all works pretty well (but needs my recent bug fix to scaffoldmaker and meshgeneratorstep).
Some code fixes needed as not happy with introducing utils/centralpath.py which is not really useful to other codes OR needs to be committed to taking utility function out of 1d path scaffold to justify its existence.
I can see a need for derivative d12 and d13 smoothing in the next step though as when you have 2 x as many elements as central path you can see their derivatives are non-smoothly interpolated. Can discuss on slack.

src/scaffoldmaker/utils/centralpath.py
elias-soltani: Thanks, Richard. This file was added before changes to meshtype_1d_path1.py. I agree it is redundant. I will remove it.

rchristie (changes requested):
2 minor fixes then it's in. Thanks.

rchristie (approved):

PR #97, opened 2020-10-09
Add path smooth cross derivatives function

rchristie:
Add edit group name option to interactive functions and enhance return values.


elias-soltani (approved):
Perfect!

PR #99, opened 2020-10-27
Parameter sets

s-fong:


src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -105,45 +166,43 @@ def generateBaseMesh(cls, region, options):
...
-                 options['Element width across arm'],
-                 options['Element thickness']]
-        elementsCount1 = options['Numbers of elements along arms']
-        elementsCount2 = 2
-        elementsCount3 = 1
-        useCrossDerivatives = False
+        parameterSetName = options['Base parameter set']
+        isDefault = 'Default' in parameterSetName
+        isMouse = 'Mouse' in parameterSetName and 'Mean' not in parameterSetName
+        isMeanMouse = 'Mean mouse' in parameterSetName
```
rchristie, line 172: As above, cleaner to have isMouse and isMean.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -166,70 +225,81 @@ def generateBaseMesh(cls, region, options):
...
+                        node = nodes.createNode(nodeIdentifier, nodetemplate)
+                        cache.setNode(node)
+                        coordinates.setNodeParameters(cache, -1, Node.VALUE_LABEL_VALUE, 1, x[ix])
+                        coordinates.setNodeParameters(cache, -1, Node.VALUE_LABEL_D_DS1, 1, ds1[ix])
+                        coordinates.setNodeParameters(cache, -1, Node.VALUE_LABEL_D_DS2, 1, ds2[ix])
+                        nodeIdentifier += 1
+                numNodesPerArm.append(len(x))
+        else:
+            numNodesPerArm = options['numNodesPerArm']
+            x_dx_all = options['meanMouseMeshNodes']
```
rchristie, line 281: Don't get these values from options. Can get these values directly since you know Base parameter set and have gleaned 'isMouse' and 'isMean'.
I think we can guarantee that the mean mouse has the same numbers of nodes, elements as the non-mean. I would prefer that the same code is used to calculate numNodesPerArm for all cases. In the generated nodes case you use len(x), but it should be computable from elementsCounts along arms.
Also rename to nodesCountsPerArm for consistency with Count being used by other variables.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -470,6 +539,7 @@ def generateBaseMesh(cls, region, options):
                 element = mesh.findElementByIdentifier(elementID)
                 markerLocation.assignMeshLocation(cache, element, addMarker["xi"])
 
+        fm.endChange()
```
rchristie, line 542: Remove beginChange/endChange as this is handled by Scaffold_base generateMesh().
I couldn't see beginChange? These must always be used in pairs, but note we don't use them any more in scaffoldmaker, instead we use the ChangeManager(fieldmodule) from opencmiss utils, which automatically calls endChange when we leave the block of code.

src/scaffoldmaker/meshtypes/meshtype_3d_stellate1.py
```diff
@@ -105,45 +166,43 @@ def generateBaseMesh(cls, region, options):
...
-        markerPoints = findOrCreateFieldNodeGroup(markerGroup, nodes).getNodesetGroup()
-        markerTemplateInternal = nodes.createNodetemplate()
-        markerTemplateInternal.defineField(markerName)
-        markerTemplateInternal.defineField(markerLocation)
+        armCount = 3
+        elementLengthCentral = options['Element width central']
+        elementLengths = [options['Element length along arm'],
+                          options['Element width across arm'],
+                          options['Element thickness']]
+        elementsCount1 = options['Numbers of elements along arms']
```
rchristie, line 190: This should be called elementsCountsAlongArms; elementsCount1 sounds like a single number.

rchristie (changes requested):
Sorry for delay, fell off my radar as we're pretty busy with deadlines.
Some fixes needed. Note I will re-review and test when these changes are done.

s-fong:
Thanks: I have made the fixes as mentioned.

rchristie (approved):

PR #108, opened 2020-11-29
Bladder updated

zekh167:
Hi Richard. I added some annotations and UBERON IDs to the bladder.


src/scaffoldmaker/annotation/bladder_terms.py
```diff
@@ -4,14 +4,16 @@
...
+    ("neck of urinary bladder", "UBERON:0001258", "FMA:15912"),
     ("Dome of the Bladder", "ILX:0738433"),
+    ("serosa of urinary bladder", "UBERON:0001260"),
+    ("bladder lumen", "UBERON:0009958"),
     ("serosa of body of urinary bladder", "ILX:0739276"),
     ("lumen of body of urinary bladder", "ILX:0739252"),
     ("serosa of neck of urinary bladder", "ILX:0739277"),
     ("lumen of neck of urinary bladder", "ILX:0739256"),
-    ("urethra", "ILX:0733022"),
+    ("urethra", "ILX:0733022", "UBERON:0000057"),
```
rchristie, line 16: Normally prefer to use the UBERON term ahead of ILX; the first one in the list after the name is always used (the rest are there in case we need to match other data). This is unless you have to match digitized contours which are annotated with the ILX term.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -806,6 +810,11 @@ def defineFaceAnnotations(cls, region, options, annotationGroups):
         lumenOfNeck = findOrCreateAnnotationGroupForTerm(annotationGroups, region, get_bladder_term("lumen of neck of urinary bladder"))
         lumenOfNeck.getMeshGroup(mesh2d).addElementsConditional(is_neck_lumen)
 
+        serosaOfurinaryBladder = findOrCreateAnnotationGroupForTerm(annotationGroups, region, get_bladder_term("serosa of urinary bladder"))
+        serosaOfurinaryBladder.getMeshGroup(mesh2d).addElementsConditional(is_urinaryBladder_serosa)
+        lumenOfurinaryBladder = findOrCreateAnnotationGroupForTerm(annotationGroups, region, get_bladder_term("bladder lumen"))
+        lumenOfurinaryBladder.getMeshGroup(mesh2d).addElementsConditional(is_urinaryBladder_lumen)
```
rchristie, line 816: Minor: capitalize start of each word in CamelCase: missed capital U in Urinary.

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -822,7 +831,7 @@ def generateUreters(region, nodes, mesh, ureterDefaultOptions,elementsCountAroun
                     elementsCountAroundUreter, trackSurfaceUreter1, ureter1Position, trackSurfaceUreter2,
                     ureter2Position, ureterElementPositionDown, ureterElementPositionAround,
                     xBladder, d1Bladder, d2Bladder, nextNodeIdentifier, nextElementIdentifier,
-                    elementsCountUreterRadial, annulusMeshGroups = []):
+                    elementsCountUreterRadial, urinaryBladderMeshGroup, annulusMeshGroups = []):
```
rchristie, line 834: First, perhaps best to rename this function to generateUreterInlets (it previously hadn't mentioned mention Ureters which was why I got Mabelle to change it, but should clarify it's just the region where they enter the bladder).
Main issue I have is that the new argument urinaryBladderMeshGroup group doesn't fit the pattern of annulusMeshGroups.
I suggest you replace both with:
bladderMeshGroups (= bladder, neck) to set on annulus and first row of ostium i.e. ostiumMeshGroups
ureterMeshGroups (= ureter, or None if not defining) to pass as vesselMeshGroups on ostium
Basically I consider the final element on the ostium as it opens into the bladder as part of the bladder, but all the elements in the ostium are in ureter group (as you can make the ostium quite long with many elements which clearly become ureters).
This may not be what you're after, for example you may want them to be only bladder or ureter, not both for the final element. See me to discuss?

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -838,7 +847,7 @@ def generateUreters(region, nodes, mesh, ureterDefaultOptions,elementsCountAroun
     nodeIdentifier, elementIdentifier, (o1_x, o1_d1, o1_d2, _, o1_NodeId, o1_Positions) = \
         generateOstiumMesh(region, ureterDefaultOptions, trackSurfaceUreter1, ureter1Position, ureter1Direction,
                            startNodeIdentifier=nextNodeIdentifier, startElementIdentifier=nextElementIdentifier,
-                           ostiumMeshGroups=None)
+                           vesselMeshGroups=[[urinaryBladderMeshGroup]], ostiumMeshGroups=[urinaryBladderMeshGroup])
```
rchristie, line 850: As above, vesselMeshGroups should only have a ureter term in them.

rchristie (changes requested):
Small changes, but one area to discuss with me in person.

src/scaffoldmaker/meshtypes/meshtype_3d_ostium1.py
```diff
@@ -725,8 +725,9 @@ def generateOstiumMesh(region, options, trackSurface, centrePosition, axis1, sta
     for v in range(vesselsCount):
         if vesselMeshGroups or ostiumMeshGroups:
             rowMeshGroups = []
-            for i in range(elementsCountAlong):
+            for i in range(elementsCountAlong - 1):
                 rowMeshGroups.append(copy.copy(vesselMeshGroups[v]) if vesselMeshGroups else [])
+            rowMeshGroups += [ostiumMeshGroups]
```
rchristie, line 730: This will have broken the heart scaffold for human and rat which use common ostium.
I think it is right to include the last bit as being both bladder + neck AND ureter; the reason is that our primary use is to project data onto, and some of the data for ureter could/should project onto these elements.

rchristie (changes requested):
Issue with ostium vessel groups.

src/scaffoldmaker/meshtypes/meshtype_3d_ostium1.py
zekh167: I think it's better to put the last layer of the ostium mesh only in one meshGroup. Since the ureters enter the bladder and make a slit on its surface. So, the elements around the ostiums are only part of the ureter and not the bladder. So, I want to set the 
 "vesselMeshGroups=[[ureterMeshGroup]]" and the "ostiumMeshGroups=[]"

rchristie: We make anatomical groups to support fitting and registration and for that purpose, the elements in the inlet zone are as much ureter as bladder. If, for example, we had data points for the outside of the ureter, some of them would need to project onto those elements, and likewise for the bladder.
I'm more concerned about there being a second row of elements that are only ureter, so the bladder has parts that are not bladder... Ideally I'd have just one element along the ostium to reduce mesh complexity, but I admit it's hard to get around the corner without distortions. I'm reasonably happy with 1 element along using:
Ostium diameter 2.2
Ostium length 0.5
Vessel end length factor 2.0
I notice the ostium has a problem with multiple elements along over such a small length due to something I've done to limit wall thicknesses, but if we decide we need 2 elements to go around this curve we'd want both to be in both bladder and ureter groups. BTW do you have left and right ureter groups?
The changes to the ostium code do break the heart scaffold & need to be undone.

rchristie (approved):
Good.

PR #111, opened 2020-12-09
Cylinder/rim

elias-soltani:
I would like to add the ability to add rim elements to the cylinder mesh, please.


src/scaffoldmaker/utils/shieldmesh.py
```diff
@@ -129,8 +130,8 @@ def getTriplePoints(self, n3):
             tx, td1 = sampleCubicHermiteCurves(
                 [self.px[n3][n2a][n1c], self.px[n3][n2c][n1b]],[[(-self.pd1[n3][n2a][n1c][c] - self.pd3[n3][n2a][n1c][c]) for c in range(3)], self.pd1[n3][n2c][n1b]],2, arcLengthDerivatives=True)[0:2]
             ltx.append(tx[1])
-            tx, td1 = sampleCubicHermiteCurves(
-                [ self.px[n3][n2a][n1b], self.px[n3][n2c][n1c] ], [ [-self.pd3[n3][n2a][n1b][c] for c in range(3)], [ (self.pd1[n3][n2c][n1c][c] + self.pd3[n3][n2c][n1c][c]) for c in range(3) ] ], 2, arcLengthDerivatives = True)[0:2]
+            # tx, td1 = sampleCubicHermiteCurves(
+            #     [ self.px[n3][n2a][n1b], self.px[n3][n2c][n1c] ], [ [-self.pd3[n3][n2a][n1b][c] for c in range(3)], [ (self.pd1[n3][n2c][n1c][c] + self.pd3[n3][n2c][n1c][c]) for c in range(3) ] ], 2, arcLengthDerivatives = True)[0:2]
```
rchristie, line 134: Why are these commented out here & in subsequent lines?

rchristie (changes requested):
It's looking great, but there are some issues.
Something in this commit has broken ventricles3 pig parameters scaffold; can you look into it?
It has some overly large d1 derivatives in the centre (and sometimes nodes around it on the minor axis direction) which causes problems, most visible if you have 8 or more elements across major and minor. (Smooth derivatives function makes them look more appropriate, but we need to get them right in the script, not rely on calling this function.)

src/scaffoldmaker/utils/shieldmesh.py
elias-soltani: I do not have points between the rim and triple points at this stage, and they will be created later on. Also, I am not using it for finding the triple point. I can remove the commented lines?

elias-soltani:
Thanks for reviewing the code all requested changes were done.

rchristie (commented):
I have approved and pulled the changes; a subsequent development will fix the element layout as discussed.

PR #112, opened 2020-12-09
Fix boundary derivative smoothing

rchristie:
Was taking derivative vector from wrong node.
Make result consistent when updating/not updating directions.


mahyar-osn (approved):
Works well!

PR #116, opened 2021-01-21
Cylinder/circular core

elias-soltani:
The core boundary (interface between core and shell) is circular (or ellipse).


rchristie (changes requested):
A couple of things to fix/discuss before getting it in.

rchristie (approved):
Looks great!

PR #122, opened 2021-02-18
Spinal cord scaffold

mahyar-osn:
The first version of the rat spinal cord  + basic annotation groups for different segments. The scaffold is a solid cylinder built based on the digitised SIMCore rate data. 

Annotation groups are proportionally created based on the 35 spinal segments:

- Cervical: 8
- Thoracic: 13
- Lumbar: 6
- Sacral: 4
- Coccygeal/Caudal: 3

All ontological ids (UBERON, ILX, and FMA) were appropriately added to the terms file. I named the terms file as `cns_terms.py` for the future to include the terms for the brainstem (and even the brain).


rchristie (commented):
I don't intend to pull this in at this time as I don't yet see its utility. I suggest retracting the PR but holding onto the code.
I think the central path should actually be owned by the whole body scaffold, but we will need to add radius variation and more annotations (which would be sufficient for tracking nerves down it). I think we need to think this through until we have a more integrated plan.

rchristie:
Closing until future plan determined.

PR #123, opened 2021-02-24
Bodycoordinates

elias-soltani:
I would like to add a new scaffold, please. The body coordinates including annotations for different sections of the body and discontinuity in d3 on the core boundary, using solid cylinder.


src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -0,0 +1,30 @@
...
+    ( "neck", "UBERON:0000974", "ILX:0733967" ),
+    ( "head", "UBERON:0000033", "ILX:0104909" ),
+    ( "skin epidermis", "UBERON:0001003", "ILX:0728574" ),
+    ( "diaphragm", "UBERON:0001103", "ILX:0103194" ),
+    ( "spinal cord", "UBERON:0002240", "ILX:0110909" ),
+    ( "body", "UBERON:0000468", "ILX:0101370" ),
+    ( "core", "" ),
+    ( "non core", "" ),
+    ( "core boundary", "" ),
+    ( "non core first layer", "" ),
```
rchristie, line 18: This 'non core first layer' should not be an annotation group. It looks like you could use a Zinc FieldGroup FieldMeshGroup independently of the Annotations to do your stuff, then remove it after use?

src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -0,0 +1,30 @@
+"""
+Common resource for body annotation terms.
+"""
+
+# convention: preferred name, preferred id, followed by any other ids and alternative names
+body_terms = [
+    ( "abdomen", "UBERON:0000916", "ILX:0725977" ),
+    ( "thorax", "ILX:0742178" ),
+    ( "neck", "UBERON:0000974", "ILX:0733967" ),
+    ( "head", "UBERON:0000033", "ILX:0104909" ),
```
rchristie, line 10: Groups head and neck don't really have the correct meaning. Perhaps these should include core and non-core in those regions, if they are needed?

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
```diff
@@ -0,0 +1,435 @@
...
+    cylinder1Settings = {
+                'Coordinate dimensions': 3,
+                'D2 derivatives': True,
+                'D3 derivatives': True,
+                'Length': 3.5,
+                'Number of elements': 1
+            }
+
+    centralPathDefaultScaffoldPackages = {
+        'Cylinder 1': ScaffoldPackage(MeshType_1d_path1, {
```
rchristie, line 41: 'Cylinder 1' is not a good name. Change to 'Default'? I initially wasn't keen on exposing this, but we could maintain some pre-fit paths for different species in future...

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
```diff
@@ -0,0 +1,435 @@
...
+            'Central path': copy.deepcopy(centralPathOption),
+            'Number of elements across major': 6,
+            'Number of elements across minor': 6,
+            'Number of elements across shell': 1,
+            'Number of elements across transition': 1,
+            'Number of elements in abdomen': 5,
+            'Number of elements in thorax': 3,
+            'Number of elements in neck': 1,
+            'Number of elements in head': 2,
+            'Shell element thickness proportion': 1.0,
```
rchristie, line 69: 'Shell element thickness proportion' is an option for the solid cylinder which I'm happy with. Here what you want is a 'Shell thickness proportion' so we can guarantee the proportion in the shell irrespective of the number of elements across it, and thus we control the body coordinates give to the core boundary. Take the parameter set here and back calculate the element thickness proportion to pass to the cylinder code.

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
```diff
@@ -0,0 +1,435 @@
...
+        middleDown = elementsCountAcrossMajor//2 - 1
+        elementIdentifiers = [elementId[1][0][middleLeft], elementId[1][topElem][middleLeft], elementId[neckFirstElem][0][middleRight], elementId[neckFirstElem][topElem][middleRight],
+                elementId[1][0][middleLeft], elementId[1][topElem][middleLeft], elementId[neckFirstElem][0][middleRight], elementId[neckFirstElem][topElem][middleRight],
+                elementId[thoraxFirstElem][middleDown][middleRight], elementId[thoraxFirstElem + 1][middleDown][middleRight], elementId[thoraxFirstElem + 1][middleDown][middleRight]]
+        markerNames = ['left hip joint', 'right hip joint', 'left shoulder joint', 'right shoulder joint',
+                       'along left femur', 'along right femur', 'along left humerus', 'along right humerus',
+                       'heart apex', 'atrial base', 'aorta']
+        locations = [[0.8, 0.5, 0.5], [0.2, 0.5, 0.5], [0.8, 0.5, 0.5], [0.2, 0.5, 0.5],
+                     [0.2, 0.99, 0.5], [0.8, 0.99, 0.5], [0.5, 0.0, 0.5], [0.5, 0.0, 0.5],
+                     [0.5, 0.5, 0.0], [0.7, 0.5, 0.0], [0.7, 0.5, 0.2]]
```
rchristie, line 348: Use a structure which groups the corresponding name, element and xi locations otherwise you create a maintenance hazard where someone can mismatch them. I suggest using a list of dict:
```
markerPoints = [
    { "name" : "left hip joint", "elementId" : elementId[1][0][middleLeft], "xi" : [0.8, 0.5, 0.5] },
    { "name" : "right hip joint", "elementId" : elementId[1][topElem][middleLeft], "xi" : [0.2, 0.5, 0.5] },
    # etc.
    ]
```
Later you can refer to them as ``markerPoints[i]["name"], markerPoints[i]["elementId"], markerPoints[i]["xi"]``.

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
```diff
@@ -0,0 +1,435 @@
...
+                ])
+        })
+    }
+
+    @staticmethod
+    def getName():
+        return '3D Whole Body 1'
+
+    @classmethod
+    def getDefaultOptions(cls, parameterSetName='Default'):
```
rchristie, line 57: I suggest you have 2 parameter sets e.g. 'Coarse' (for the current default values) and 'Fine' (for planned higher resolution model we will fit with) and make 'Default' (which all scaffolds support) give the same results as Coarse (but we could change it to match Fine in future).

rchristie (changes requested):
Very good... some small fixes, a couple which we should discuss.
I consider the marker points a work-in-progress which we will change when we decide on a more final mesh resolution in future fits. I am hoping that the limbs can be oriented radially on the body, for example. I want this scaffold in as soon as possible & we can adjust things in future PRs.

src/scaffoldmaker/annotation/body_terms.py
elias-soltani: Done

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
elias-soltani: Done

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
elias-soltani: Thanks for the suggestion, it makes total sense to use this.

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
elias-soltani: Done

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody1.py
elias-soltani: The cylinder mesh only allows the same 'Shell element thickness proportion' in major and minor directions. We can either have the same 'shell element thickness proportion' or the same 'sell thickness proportion' in major and minor directions if the number of elements across major and minor is not the same. If these are the same then both can happen at the same time.

elias-soltani:
Thank you very much for reviewing the code. I changed it according to the suggestions.

src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -8,6 +8,8 @@
     ( "thorax", "ILX:0742178" ),
```
rchristie, line 8: I looked up thorax and abdomen in SciCrunch and these include the the outer tissue and skin layers so are not appropriate terms. I'm thinking we need to change them to be:
'thoracic cavity'
'abdominal cavity'/'pelvic cavity' or combined 'abdominopelvic cavity'.
If so, we need to make the spinal cord slightly outside.
Alternatively we must define our own terms representing what we mean by the core.

rchristie (changes requested):

rchristie (approved):
Good.
Still thinking of making abdomen, thorax, neck and head go right through, but can do later.

elias-soltani:
Thanks for reviewing the code and great suggestions.

PR #126, opened 2021-03-22
Fix unused scale factor bugs

rchristie:
Fix lung apex annotation.

### DIFF
--- a/src/scaffoldmaker/utils/annulusmesh.py
+++ b/src/scaffoldmaker/utils/annulusmesh.py
@@ -412,7 +412,7 @@ def createAnnulusMesh3d(nodes, mesh, nextNodeIdentifier, nextElementIdentifier,
                 scaleMinus1 = mapStartLinearDerivativeXi3 or mapEndLinearDerivativeXi3
                 if (not scaleMinus1) and mapStartDerivatives:
                     for n3 in range(2):
-                        # need to handle 3 or 4 maps (e1 uses first 3, en uses last 3)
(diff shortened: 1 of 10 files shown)


src/scaffoldmaker/utils/annulusmesh.py
```diff
@@ -410,22 +410,28 @@ def createAnnulusMesh3d(nodes, mesh, nextNodeIdentifier, nextElementIdentifier,
...
+                if (not scaleMinus1) and mapStartDerivatives:
                     for n3 in range(2):
-                        for i in range(2):
-                            derivativesMap = startDerivativesMap[n3][e1] if (i == 0) else startDerivativesMap[n3][en]
-                            for map in derivativesMap:
-                                if map and (-1 in map):
-                                    scaleMinus1 = True
-                                    break
-                if (not scaleMinus1) and endDerivativesMap:
+                        # need to handle 3 or 4 maps (e1 uses first 3, en uses last 3)
```
mlin865, line 415: Correct comment - should say "e1 uses last 3, en uses first 3"

mlin865 (approved):
Looks good to me.  Just a minor revision for the comment in annulus and it is good to go.

src/scaffoldmaker/utils/annulusmesh.py
rchristie: Fixed.

arti-sukasem (approved):

PR #128, opened 2021-03-28
Brainstem

s-fong:


src/scaffoldmaker/scaffolds.py
```diff
@@ -88,6 +90,7 @@ def __init__(self):
             MeshType_3d_stomach1,
             MeshType_3d_stomachhuman1,
             MeshType_3d_tube1,
+            MeshType_3d_tubeseptum1
```
rchristie, line 93: Remove this line which has crept in.

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
```diff
@@ -0,0 +1,355 @@
...
+        faceGroup.getMeshGroup(mesh2d).addElementsConditional(is_face_ext)
+
+        # external regions
+        namelist = ['midbrain', 'pons', 'medulla oblongata']
+        for subregion in namelist:
+            subGroup = AnnotationGroup(region, get_brainstem_annotation_term(subregion))
+            issub = subGroup.getFieldElementGroup(mesh2d)
+            is_subface = fm.createFieldOr(fm.createFieldAnd(issub, is_exterior_face_xi1), fm.createFieldAnd(issub, is_exterior_face_xi3))
+            subFaceGroup = findOrCreateAnnotationGroupForTerm(annotationGroups, region, (subregion+'_exterior', None))
+            subFaceGroup.getMeshGroup(mesh2d).addElementsConditional(is_subface)
```
rchristie, line 355: Always end files with new line character - can see the warning symbol.
It's a coding standard thing.

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
```diff
@@ -0,0 +1,355 @@
...
+"""
+
+from __future__ import division
+import copy
+from opencmiss.zinc.element import Element
+from opencmiss.zinc.node import Node
+from opencmiss.utils.zinc.field import Field, findOrCreateFieldCoordinates, findOrCreateFieldGroup, findOrCreateFieldNodeGroup, findOrCreateFieldStoredMeshLocation, findOrCreateFieldStoredString
+from opencmiss.utils.zinc.finiteelement import getMaximumNodeIdentifier
+from scaffoldmaker.annotation.annotationgroup import AnnotationGroup, findOrCreateAnnotationGroupForTerm
+from scaffoldmaker.annotation.brainstem_annotation_terms import get_brainstem_annotation_term
```
rchristie, line 12: You haven't added this new file; please also rename to brainstem_terms, get_brainstem_term, for consistency with other scaffolds.

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
```diff
@@ -0,0 +1,355 @@
...
+from scaffoldmaker.meshtypes.meshtype_1d_path1 import MeshType_1d_path1
+from scaffoldmaker.utils.meshrefinement import MeshRefinement
+from scaffoldmaker.utils.cylindermesh import CylinderMesh, CylinderShape, CylinderEnds, Tapered, ConeBaseProgression, CylinderCentralPath
+from scaffoldmaker.utils.zinc_utils import exnodeStringFromNodeValues
+from scaffoldmaker.scaffoldpackage import ScaffoldPackage
+
+
+class MeshType_3d_brainstem1(Scaffold_base):
+    """
+Generates a tapered cylinder for the brainstem based on solid cylinder mesh, with variable numbers of elements in major, minor and length directions. Regions of the brainstem are annotated.
```
rchristie, line 23: Indent to level of """

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
```diff
@@ -0,0 +1,355 @@
...
+            'Number of elements across minor': 6,
+            'Number of elements along': 8,
+            'Lower half': False,
+            'Use cross derivatives': False,
+            'Refine': False,
+            'Refine number of elements across major and minor': 1,
+            'Refine number of elements along': 1,
+            'Taper major increment': 0.2,
+            'Taper minor increment': 0.1,
+            'Along length': 3
```
rchristie, line 66: Move these arguments above Use cross derivatives (our style is to have the sub-scaffolds and number of elements, first, refine parameters last, others nominally alphabetical).
Rename 'Along length' to just 'Length' I think.

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
```diff
@@ -0,0 +1,355 @@
...
+        ############################################################
+        eIndexPM = {}
+        xiPM = {}
+        pointMarkers = {}
+        eIndexPM['caudal-dorsal'] = int(elemPerLayer/2)
+        eIndexPM['midRostCaud-dorsal'] = int(elemPerLayer / 2) + (elemPerLayer * int((elementsCountAlong/2)-1))
+        eIndexPM['rostral-dorsal'] = (elemPerLayer*(elementsCountAlong-1)) + int(elemPerLayer/2)
+        eIndexPM['caudal-ventral'] = int(elemPerLayer/2) - (elementsCountAcrossMinor-1)
+        eIndexPM['midRostCaud-ventral'] = eIndexPM['midRostCaud-dorsal'] - (elementsCountAcrossMinor-1)
+        eIndexPM['rostral-ventral'] = int((elemPerLayer*(elementsCountAlong-1)) + (int(elemPerLayer/2) - elementsCountAcrossMinor + 1))
```
rchristie, line 236: We're now adding the marker points as annotation groups and putting the points in not just the marker group, but the brainstem group and their own annotation group.
We will add the point annotations with the name and term Id "None" (a string with the word None in it) and request the terms from SciCrunch using the google sheet the mappers maintain.
All the names need to be qualified with "brainstem" and point/junction so they are clear what anatomy they are for in SciCrunch. Since I can't run it yet, I can't comment any further on the locations. Avoid abbreviations like midRostCaud.

rchristie (changes requested):
The terms file is missing (please also rename - see comment) and a syntax error meaning I can't run it to check. Please check it is working for you.
I realise you're moving on so some of these task may not need to be done by you (may be done later), but it would be good to get it running properly and in the codebase.

s-fong:
I have pushed these changes. The mesh file I have added includes updated markers for where cranial nerves 3-12 emerge from the brainstem, which was not considered at the time of this original pull request.

s-fong:
Sorry, I have pushed the wrong version up. Please don't review this one yet, it won't work!

s-fong:
This should work for you now. However, cranial nerve emergent markers are not included as they are computed in "brainstem_coordinates" which are not defined on this mesh.

src/scaffoldmaker/annotation/brainstem_terms.py
```diff
@@ -0,0 +1,23 @@
...
+Common resource for testing annotation terms.
+"""
+
+# convention: preferred name, preferred id, followed by any other ids and alternative names
+brainstem_terms = [
+                    ("medulla oblongata", "FMA:62004", "UBERON:0001896"),
+                    ("pons", "FMA:67943", "UBERON:0000988"),
+                    ("midbrain", "FMA:61993", "UBERON:0001891"),
+                    ("diencephalon", "FMA:62001", "UBERON:0001894"),
+                    ("brainstem", "FMA:79876", "UBERON:0002298")
```
rchristie, line 11: The preferred IDs are the UBERON ones, which should be second; don't need the FMA ID column.

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
```diff
@@ -0,0 +1,376 @@
...
+        eIndexPM['caudal-ventral'] = int(elemPerLayer/2) - (elementsCountAcrossMinor-1)
+        eIndexPM['midRostralCaudal-ventral'] = eIndexPM['midRostralCaudal-dorsal'] - (elementsCountAcrossMinor-1)
+        eIndexPM['rostral-ventral'] = int((elemPerLayer*(elementsCountAlong-1)) + (int(elemPerLayer/2) - elementsCountAcrossMinor + 1))
+        xiPM['caudal-ventral'] = [1.0, 0.0, 0.0]
+        xiPM['caudal-dorsal'] = [1.0, 0.0, 1.0]
+        xiPM['midRostralCaudal-ventral'] = [1.0, 1.0, 0.0]
+        xiPM['midRostralCaudal-dorsal'] = [1.0, 1.0, 1.0]
+        xiPM['rostral-ventral'] = [1.0, 1.0, 0.0]
+        xiPM['rostral-dorsal'] = [1.0, 1.0, 1.0]
+        for key in eIndexPM.keys():
```
rchristie, line 229: We're making the marker points as annotationGroups now. This makes the code cleaner as there isn't the risk of not matching by name, however it means we must request ILX terms from K-core; mainly requires putting the name and description on a Google spreadsheet.
These names are not sufficient however, as they must be qualified with 'Brainstem', 'midline' and 'point' to be fully prescriptive. Is 'rostral' or 'cranial' more appropriate here -- we need terms that make sense across species.
The names should be plain english, uncapitalised without camelCase.
I don't think we need triple qualified rostral+caudal+ventral/dorsal as only 2 should be sufficient (along with midline).
Wouldn't the midRostralCaudal* points be better on the medulla-pons boundary?
Alternatively to many of these points, we could define a midline group around the boundary...

rchristie (changes requested):
Some feedback. We need to have a discussion/handover.
I'd like to get some idea of the sizes and shapes in different species.
Is the midbrain normally longer than that?
Can we make annotation groups for the caudal end where it attaches to the spinal cord, and the cranial end?

src/scaffoldmaker/meshtypes/meshtype_3d_brainstem.py
s-fong: I am unsure what you want here. I think it's probably better to talk to the next mapper to make these changes.

rchristie (approved):
Fiducial marker points need to be redefined as annotation groups and qualified with brainstem and midline (or similar).

PR #132, opened 2021-04-21
Stomach

mlin865:


src/scaffoldmaker/meshtypes/meshtype_3d_stomach1.py
```diff
@@ -0,0 +1,2473 @@
...
+            'Number of elements between annulus and duodenum': 6,
+            'Number of elements through wall': 1,
+            'Number of radial elements in annulus': 1, 
+            'Wall thickness': 5.0,
+            'Limiting ridge': False,
+            'Gastro-esophagal junction': copy.deepcopy(ostiumOption),
+            'Gastro-esophagal junction position along factor': 0.35,
+            'Annulus derivative factor': 1.0,
+            'Use cross derivatives': False,
+            'Use linear through wall' : False,
```
rchristie, line 253: When I use this option I get:
AssertionError: eftfactory_bicubichermitelinear.createEftWedgeCollapseXi1Quadrant:  Failed to validate eft
It even crashed when tried with multiple elements through the wall ... investigate.

src/scaffoldmaker/meshtypes/meshtype_3d_stomach1.py
```diff
@@ -0,0 +1,2473 @@
...
+        else:
+            centralPathOption = cls.centralPathDefaultScaffoldPackages['Human 1']
+            ostiumOption = cls.ostiumDefaultScaffoldPackages['Human 1']
+
+        options = {
+            'Central path': copy.deepcopy(centralPathOption),
+            'Number of elements around esophagus': 8,
+            'Number of elements around duodenum': 12,
+            'Number of elements between annulus and duodenum': 6,
+            'Number of elements through wall': 1,
```
rchristie, line 245: The annulus is not also supporting multiple elements through the wall. Do you want to update it to do so?
May be best to remove it until you're ready. There are issues of transitioning to fewer elements at each end which will be tricky.

src/scaffoldmaker/meshtypes/meshtype_3d_stomach1.py
```diff
@@ -0,0 +1,2473 @@
...
+            'Wall thickness': 5.0,
+            'Limiting ridge': False,
+            'Gastro-esophagal junction': copy.deepcopy(ostiumOption),
+            'Gastro-esophagal junction position along factor': 0.35,
+            'Annulus derivative factor': 1.0,
+            'Use cross derivatives': False,
+            'Use linear through wall' : False,
+            'Refine': False,
+            'Refine number of elements around': 1,
+            'Refine number of elements along': 1,
```
rchristie, line 256: Merge these to ``Refine number of elements surface`` as they must be the same when you have tricky general linear maps.

src/scaffoldmaker/meshtypes/meshtype_3d_stomach1.py
```diff
@@ -0,0 +1,2473 @@
...
+            centralPathOption = cls.centralPathDefaultScaffoldPackages['Human 1']
+            ostiumOption = cls.ostiumDefaultScaffoldPackages['Human 1']
+
+        options = {
+            'Central path': copy.deepcopy(centralPathOption),
+            'Number of elements around esophagus': 8,
+            'Number of elements around duodenum': 12,
+            'Number of elements between annulus and duodenum': 6,
+            'Number of elements through wall': 1,
+            'Number of radial elements in annulus': 1,
```
rchristie, line 246: When set to 2 this does some funny stuff on the lesser curvature for the rat ... can you look into. May be an issue in annulus mesh which can be fixed later. May want to add an issue on github for it?
Should this be ``Number of elements across cardia``? User doesn't know which annulus you're referring to.

rchristie (changes requested):
Very good, but some fixes needed as in review comments.
Note the human stomach has uneven bunching up of elements on the lesser curvature. Rat stomach has some nodes in non-ideal positions on cardia limiting ridge adjacent to 6-way point on lesser curvature. I'll accept this PR without these fixes, but we need to fix them soon.
Note when I properly smoothed the human central path (as requested in review comment), the bunching was almost gone, so this may be enough.

src/scaffoldmaker/annotation/stomach_terms.py
```diff
@@ -0,0 +1,30 @@
...
+# convention: preferred name, preferred id, followed by any other ids and alternative names
+stomach_terms = [
+    ( "body of stomach", "UBERON:0001161", " FMA:14560", "ILX:0724929"),
+    ( "cardia of stomach", "UBERON:0001162", " FMA:14561", "ILX:0729096"),
+    ( "duodenum", "UBERON:0002114", " FMA:7206", "ILX:0726125"),
+    ( "esophagus", "UBERON:0001043", "FMA: 7131", "ILX:0735017"),
+    ( "esophagogastric junction", "UBERON:0007650", "FMA: 9434", "ILX:0733910"),
+    ( "forestomach-glandular stomach junction", "UBERON:0012270", "ILX:0729974"),
+    ( "forestomach-glandular stomach junction on inner wall", None),
+    ( "forestomach-glandular stomach junction on outer wall", None),
```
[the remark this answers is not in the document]
mlin865: Suggest we keep the terms as-is now but I will update the terminologies and submit the new terms for SAWG review after discussion with the investigators.

src/scaffoldmaker/meshtypes/meshtype_3d_stomach1.py
mlin865: Could be due to the missing "setEftScaleFactorIds(eft, [1], []) " in a section which requires scale factor -1. I have updated eftfactory_bicubichermitelinear.createEftWedgeCollapseXi1Quadrant to reflect that. I am able to generate the the scaffold with linear through wall with multiple elements through wall. Can you please try on your mchine and let me know if this is not resolved?

rchristie (approved):
Only minor comment is whether you need the annotation group "esophagogastric junction".
Great work!

PR #139, opened 2021-05-13
Added Rat lung mesh parameters

arti-sukasem:


rchristie (approved):
There are some minor things that you can fix later. The dorsal 'box' part of the rat accessory lobe should rise off the diaphragm like the mouse I think. Also, I'd like the laterodorsal point on the middle right mouse lobe defined on mouse and rat too; can you check whether that makes sense?

arti-sukasem:
I had a look at the dataset. As you can see in the attachments, the dorsal side of the right and left lung in the rat dataset are at the same level. You're also right about the rise-off the diaphragm; the mesh already showed a rise-off at the dorsal block.

Screenshot 2021-05-19 220600

Screenshot 2021-05-19 220600

PR #149, opened 2021-07-02
Mouse brainstem

arti-sukasem:
I edited the mouse brainstem that we discussed today. I made the midbrain looks bigger than the previous version that I showed you. I think this will eliminate the distortion in the midbrain dataset. Thank you, Richard.


rchristie (approved):
Great.

PR #150, opened 2021-07-04
Pig lung

arti-sukasem:
I added the pig lung to the human lung section and added the right lung accessory lobe to the pig lung using 'if isPig'. I also separated the pig lung annotation from the human lung annotation.


arti-sukasem:
it moves to PigLung_1 branch #175

PR #153, opened 2021-07-16
Add parameter sets for mouse and pig bladder

keeran97:


src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -46,11 +46,45 @@ class MeshType_3d_bladderurethra1(Scaffold_base):
                     [[0.0, 0.0, 15.0], [0.0, 0.0, 15.0], [0.0, 0.5, 0.0], [0.0, 0.0, -0.5]],
                     [[0.0, 0.0, 30.0], [0.0, 0.0, 15.0], [0.0, 0.5, 0.0], [0.0, 0.0, -0.5]],
                     [[0.0, 0.0, 45.0], [0.0, 0.0, 15.0], [0.0, 0.5, 0.0], [0.0, 0.0, -0.5]],
-                    [[0.13, 0.1, 60.45], [0.4, 1.1, 13.5], [0.0, 0.5, 0.0], [0.0, 0.0, -0.5]],
```
rchristie, line 49: All central paths here have small components in the x direction which make them asymmetric - these components should be zero i.e. curves only in y-z plane.
Can you please make sure that mouse, pig have zero components for all x components of VALUE, D_DS1, D_DS2 then in meshgenerator smooth derivatives, make sides normal and smooth side cross derivatives then list parameters with appropriate precision (say 4-5 sig figs for largest value) then re-paste in here. Ideally all the other species should also be fixed... to discuss.
[Another comment is that this is the only organ that doesn't give the models an absolute scale in mm for each species; they are all about the same size; I'm not asking you to change this yet, just a comment for now.]

src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -281,6 +412,30 @@ def getDefaultOptions(cls, parameterSetName='Default'):
...
+            options['Minor diameter'] = 30.0
+            options['Wall thickness'] = 0.3
+            options['Neck diameter 1'] = 8.0
+            options['Neck diameter 2'] = 7.5
+            options['Neck angle degrees'] = 25
+            options['Ureter position around'] = 0.67  # should be on the dorsal part (> 0.5)
+            options['Ureter position down'] = 0.83
+            options['Urethra diameter 1'] = 2.0
+            options['Urethra diameter 2'] = 2.0
+            options['Urethra wall thickness'] = 0.25
```
rchristie, line 438: Mouse, pig have followed cat in being quite inflated, rat is half inflated, human perhaps a little more. Need to consistently choose a state, probably partly inflated so regular shaped, not over inflated so strained, not under-inflated so wrinkled up, to be the generic reference state. Do a bit of research and propose what you think is the best state.

rchristie (changes requested):
Please catch up to latest scaffoldmaker and meshgenerator step so we can see if working with latest changes.
I think all these bladder meshes have too many elements, particularly up the length. Should have fewer like human bladder.

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -156,7 +156,7 @@ def warpSegmentPoints(xList, d1List, d2List, segmentAxis,
                 vectorToFirstNode = [xRot1[c] - centroidRot[c] for c in range(3)]
                 if vector.magnitude(vectorToFirstNode) > 0.0:
                     cp = vector.crossproduct3(vector.normalise(vectorToFirstNode), sd2[nAlongSegment])
-                    if vector.magnitude(cp) > 0:
+                    if vector.magnitude(cp) > 1e-12:
```
rchristie, line 159: Don't hardcode magic numbers such as these in reusable utility library code. It will not work for all scales of models. Can you determine a tolerance based on ranges of coordinates passed in.

rchristie (changes requested):
I did not expect you to change the number of elements along the central path. This doesn't really affect the results. It's ok as 8 was probably too many.
I was wanting you to change the 'Number of elements along bladder' for the bladder itself to be less than 12, hopefully 8 like the human.
I'm not sure it's the right time to change the existing bladders, or make all of them actual size. I'd suggest going back to making them about the same size and just customise shape of mouse, pig?
(In any case, mouse is the same size as rat -- I'd expect smaller.)

rchristie (approved):
Good.

PR #154, opened 2021-08-27
added pig and sheep brainstem parameters

arti-sukasem:
I added two brainstem scaffolds (i.e. pig and sheep). I have datasets with me. please let me know if you need them, I will send them to you right away :)


arti-sukasem:
this moves to PR #176 - updates the branch due to previous map client updates

PR #157, opened 2021-10-13
Colon coordinates

mlin865:


rchristie (changes requested):
Colon coordinates are looking nicely round except for cattle which seems a little flat at it's single tenia coli (I wasn't aware it only had one and it doesn't protrude on the serosa either - is that correct?) Perhaps need more elements around cow as the aspect ratio difference is extreme by the tenia.
The layer thicknesses should be the same for all species in *colon coordinates* (but can be different in geometric *coordinates* because we are linear through the wall), otherwise they are not comparable. The longitudinal layer appears to be thinner on average, but the others more similar. Could either do 0.25 for all, or some compromise between species e.g. 0.35 0.25 0.25 0.15. I can't see a problem with using 0.25 for all; I see you've even appropriately chosen derivatives through the wall if you turn on hermite -- nice!
I think the tenia thickness should be greater in colon coordinates -- it tends to be thicker than the longitudinal muscle layer - want a balance between species, or match the consistent 0.25 of wall thickness.
We still need to be able to create a single layer scaffold as before, but when I enter a number other than 4 it is set back to 4. At a minimum, 1 needs to be allowed. Idea: if we want to support 2 or 3 layers e.g. if mucosa and submucosa were combined into just 'mucosa', we could set the relative thickness of submucosa to 0.0 to ignore, etc.?
I would say it's only right to define the 'myenteric nerve plexus' as a 2D layer if you expected the nerve data to be projected onto that 2D layer -- I would have thought the nerve data had a small depth in and out of the neighbouring muscle layers.
That said, to fit the colon or part of it to data with distinct layers, we'll want 2D annotation groups for all in-between layers. I see we have the 2D 'serosa' on the outside, but there isn't a corresponding 2D group for the inside surface. That would allow fitting the boundary between the 2 muscle layers so we can embed nerve data in the plexus region overlapping both layers...

rchristie (approved):
Great work - amazing model.

PR #158, opened 2021-10-19
Add a one-headed muscle type scaffold

elias-soltani:
I would like to add an initial one-headed muscle type scaffold.


elias-soltani:
Another PR was created for this one.

PR #164, opened 2021-10-30
Import optimisations and PEP8 conformance changes

hsorby:


rchristie (changes requested):
Lot's of great improvements here.
However, I don't want the parameter lists to be reformatted -- it's helpful that they are fixed format, and they will be generated that way if ever changed so we'll see what is different.
They are also large chunks to change in the history and I think it's important for the code lines to credited/blamed to the scaffolders who put them in.

PR #170, opened 2021-12-06
Created a 3D bifurcation tree handling rotations/twists

arti-sukasem:
Hi Richard, 
I created a new pull request for the airway tree, since the previous pull request was not compatible with the new mapclient.
I also updated the code, which allows twists and rotations along a 1D bifurcation tree. 
Thank you


rchristie (changes requested):
Firstly, the mesh looks pretty nice.
I didn't quite get to the main algorithm but have a fair number of issues already -- see comments in code.
For clarity, the main algorithm needs to be broken up into functions which do key jobs (with nice names so it's clear what they do), and in particular recursive functions for adding nodes and elements for each section, which then propagate to its children. Note we can renumber nodes and elements afterwards to be by generation, if we want.
I'm expecting once you've rationalised the algorithm it will be much clearer for me to review.
After this the mesh is still not usable as vessels interpenetrate at early junctions in the tree. This will require different junction meshes which can be added later.

PR #174, opened 2022-02-03
add source coordinate name as input argument

mahyar-osn:
A tiny change to the `MeshRefinement` class to accept a name for the `sourceCoordinates`. This makes it more flexible when we have a mesh with a different field name than the default `coordinates`.


rchristie:
What we need to do is refine all source fields to target fields of the same name, now that we are having various organ-specific coordinates in addition to geometric coordinates. Note what you've currently done will make a target field called coordinates regardless of the name of the incoming field.

PR #179, opened 2022-03-10
Adjust range of elements according to updated number of elements.

elias-soltani:


rchristie (changes requested):
Can you change these options to "Crop elements in direction X" with values 0, 0 normally -- the first value crops from the low side, the last from the high side.
Storing additional parameters such as "Current number of elements" is a last resort -- I only use this for remembering the parameters set in certain cases.
I notice with the current implementation you can't change the first value from 0 anyway.
Can you please also catch this branch up to main as it's missing the supporting functions for user-defined marker points.
You should update your mesh generator step too: note it has been renamed Scaffold Creator / mapclientplugins.scaffoldcreator as part of the changes for documentation. A new mapclient will migrate workflows to use it.

elias-soltani:
Changes implemented.

rchristie (approved):

PR #184, opened 2022-03-19
Add user marker point annotation groups

rchristie:
Use simpler marker annotation API in brainstem1 and heartventricles1.


hsorby (approved):
Apart from the zinc_utils.py part.

PR #194, opened 2022-05-11
Generic lung

arti-sukasem:
Create a generic lung scaffold for human, mouse, rat, pig with its geometric and material coordinates.


src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+                fm_n1 = region_n1.getFieldmodule()
+                coordinates_n1 = findOrCreateFieldCoordinates(fm_n1, name="lung coordinates")
+                # point reference
+                fm = fm_n1
+                coordinates = coordinates_n1
+                # update material coordinates according to geometric parameters
+                options['Material Parameters']['Number of left lung lobes'] = options['Number of left lung lobes']
+                options['Material Parameters']['Open fissures'] = options['Open fissures']
+                options['Material Parameters']['Accessory lobe'] = options['Accessory lobe']
+                options = options['Material Parameters']
```
rchristie, line 294: A simpler and more efficient way to do this is to make the "coordinates" in the shape of the material coordinates first (since your transformations start from that shape), then do the following:

1. temporarily rename "coordinates" to "lung coordinates" (requires lung coordinates to not be defined yet).
   `coordinates.setName("lung coordinates")`
2. write the model to memory
   `# set up sir with a memory resource then:`
   `region.write(sir)`
3. Rename coordinates back:
   `coordinates.setName("coordinates")`
4. read the memory buffer
   `# set up sir2 with a memory buffer resource as written earlier:`
   `region.read(sir2)`
5. Find material coordinate field if needed later, e.g. as part of your functions:
   `material_coordinates = fm.findFieldByName("lung coordinates").castFiniteElement()`
6.  Now apply your transformation functions to the coordinates field only.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Accessory lobe dorsal width': 0.5,
+            'Accessory lobe ventral height': 0.5,
+            'Accessory lobe ventral width': 0.5,
+            'Diaphragm centre x': 0.0,
+            'Diaphragm centre y': 0.0,
+            'Diaphragm curvature x': 0.0,
+            'Diaphragm curvature y': 0.0,
+            'Rotation in degree about [z, y, x]: Left lung': [0.0, 0.0, 0.0],
+            'Rotation in degree about [z, y, x]: Right lung': [0.0, 0.0, 0.0],
+            'Rotation in degree about [z, y, x]: Accessory lobe': [0.0, 0.0, 0.0],
```
rchristie, line 90: "Left lung rotation in degrees about z, y, z" etc. following wording for the general rotation parameter, and also put options in with the options for Left, Right and Accessory lobe above.
See comments later about these being applied in the opposite sense to that expected (see what the general rotation parameter does with a positive rotation, it's a positive right hand rotation about the respective axis).
It seems these rotations are about the origin (see comments with the function later).
Will it ever make sense to rotate about y or x? Rotations on those axes affects the curve of the diaphragm...

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+
+    @classmethod
+    def getDefaultOptions(cls, parameterSetName='Default'):
+        if parameterSetName == 'Default':
+            parameterSetName = 'Human 1'
+        options = {
+            'Base parameter set': parameterSetName,
+            'Number of left lung lobes': 2,
+            'Left-right lung spacing': 1.0,
+            'Left-right apex closing space': 0.0,
```
rchristie, line 61: Rename "closing space" to "medial shear" which is what it is.
If you put a value 1.0 in, it shears by 0.5 units at the apex; it would be more logical for it to move 1.0 units in the medial direction in this case.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Left-right apex along y-axis': 0.0,
```
rchristie, line 62: This is also a shear in the -y direction, so make it "Left-right apex dorsal shear" and similarly fix the units.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Left lung width': 0.25,
+            'Left lung depth': 0.5,
```
rchristie, line 64: Left and Right lung width, depth should be twice this. Users understand whole dimensions like diameters, but we coders should convert them to radius or major/minor axis lengths by dividing by 2 when we use them.
Height is fine since we don’t include the bottom half of the ellipsoid.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Left lung height': 1.0,
+            'Left lung ventral edge sharpness factor': 0.0,
```
rchristie, line 66: The sharpness factor 1.0 used to make the front razer sharp, now it's about 45 degrees. Not sure why?

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Left lung medial curvature factor': 0.0,
+            'Left lung medial protrusion factor': 0.0,
```
rchristie, line 68: This does nothing at value 0.0, works well at 1.0, but values like 0.1 are not 10% of the position between 0.0 and 1.0 as expected, instead it makes the whole lung much thinner. This needs looking into.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Right lung width': 0.25,
```
rchristie, line 69: Same issues for Right as mentioned for left.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            'Accessory lobe ventral height',
+            'Accessory lobe ventral width',
+            'Accessory lobe medial curvature',
+            'Diaphragm centre x',
+            'Diaphragm centre y',
+            'Diaphragm curvature x',
+            'Diaphragm curvature y',
+            'Rotation in degree about [z, y, x]: Left lung',
+            'Rotation in degree about [z, y, x]: Right lung',
+            'Rotation in degree about [z, y, x]: Accessory lobe',
```
rchristie, line 243: Move these options up with the Left, Right, Accessory options.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+            options['Left lung width'] = 0.25
+            options['Left lung depth'] = 0.7
+            options['Left lung height'] = 1.0
+            options['Left lung ventral edge sharpness factor'] = 0.6
+            options['Left lung medial curvature factor'] = 1.0
+            options['Left lung medial protrusion factor'] = 0.8
+            options['Right lung width'] = 0.18
+            options['Right lung depth'] = 0.9
+            options['Right lung height'] = 1.0
+            options['Right lung ventral edge sharpness factor'] = 0.8
```
rchristie, line 160: The Right and (to a lesser extent Left) sharpness is too great on the rat -- I think you should assume the rat data is quite deflated so choose options to inflate it a bit, so more similar to the state of the other species, particularly the mouse which I'd think it would be similar to.
I notice the pig has even higher sharpness factors; I'm not sure what other options cause it to not look as sharp as the rat here?

rchristie (changes requested):
In general, the lungs are looking really good and I want them in the code! I'm super pleased how the few parameters for the accessory lobe make it look pretty good.
There are some specific things to fix listed in the code.
Also add a lung2 test to test_lung.py. Get numbers of annotation groups etc., and check a selection of sizes perhaps.

With the above fixed I'm very happy to pull the code in, but there are some follow up things to do:
1. Think about making the horizontal fissure straight in material coordinates, so it's easy to describe. Ideally, all divisions between regions in the material coordinates should be simple to describe functions i.e. not "whatever my scaffold made". Want to generally put some further thinking into the choice of material coordinates everywhere, before they are fully used by others.
2. A known issue is that mesh refinement closes up the open fissures; we need to add features (and probably supply extra arguments) to ensure nodes on the wrong side of a fissure are not able to used by elements. We'll have to think about this. Other issues are that we should do a better job of refining triangles, and we need to support defining other fields such as the material coordinates on the refined mesh. Keep these issues in mind.
3. Need to show Merryn/lung group and get their approval too, particularly about material coordinates.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
rchristie: I think you should try having a vertical "curvature factor" which brings the top inward as bending seems a more smooth deformation than shear (which makes derivatives less normal to each other).
While I'm think about it, your options 'Left/Right lung medial curvature factor' need to specify that this is about the z axis, and the new curvature I'm talking about here is about the x axis, but a positive curvature is also in the medial direction. Maybe z is "depth" and x is "height", so depth medial curvature and height medial curvature.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
```diff
@@ -0,0 +1,2139 @@
...
+                fissureGroup = fm.createFieldAnd(lobe["upper lobe of left lung"], lobe["lower lobe of left lung"])
+            elif fissureTerm == "oblique fissure of right lung":
+                fissureGroup = fm.createFieldAnd(fm.createFieldOr(lobe["middle lobe of right lung"], lobe["upper lobe of right lung"]),
+                                                        lobe["lower lobe of right lung"])
+            elif fissureTerm == "horizontal fissure of right lung":
+                fissureGroup = fm.createFieldAnd(lobe["upper lobe of right lung"], lobe["middle lobe of right lung"])
+            fissureSurfaceGroup = findOrCreateAnnotationGroupForTerm(annotationGroups, region, get_lung_term(fissureTerm))
+            fissureSurfaceGroup.getMeshGroup(mesh2d).addElementsConditional(fissureGroup)
+
+def createLungNodes(spaceFromCentre, lengthUnit, widthUnit, heightUnit, fissureAngle, obliqueProportion,
```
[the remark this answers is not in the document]
arti-sukasem: No. The following functions are not shared with lung1. They are completely different since lung2 generates nodes from the defined parameters while Lung1 is much simpler creating nodes without any parameters.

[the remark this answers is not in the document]
rchristie: That is true for node parameters.
However, I often separate code into 3 stages: making parameter arrays, building nodes, building elements, and the latter 2 steps should be the same (for the case without open fissures).
I'm not going to press for this now unless you can see obvious common functions.
Generally, unless code is trivial or the differences are throughout a piece of code (turning it into "spaghetti", we do our best to put it into common functions when possible.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
arti-sukasem: I fixed the parameters for the rat. This is because of the width of the right lung which causes the right lung is even more sensitive to its sharpness parameter

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
arti-sukasem: I changed the units to be in degree

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
arti-sukasem: Okay I set the maximum sharpness factor to be one (thinnest as it could get in the ventral). However, if it goes beyond 1, it will cause an element line crossing each other. I didn't know if we need to thinen the ventral end that much.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
arti-sukasem: which parameter are you talking about? sharpness factor?

rchristie: Protrusion factor.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
arti-sukasem: I removed rotations about y and x-axis as you suggested and make it apply to a right-hand rule.
I also fixed the rotation at the centre of elements.

src/scaffoldmaker/meshtypes/meshtype_3d_lung2.py
arti-sukasem: I ended up having parameterised scaffold in the 'material coordinates'. I am not sure if this is what you are looking for

PR #195, opened 2022-05-11
Update heart terms, improve epicardium, add docs

rchristie:
Shift fat pad nodes around inlets to reduce distortion.
Update heart terms to have luminal/outer surfaces for digitising.
Use preferred term name and ids.


mahyar-osn (approved):
I read every line and tried to be super careful. All look good to me.

PR #197, opened 2022-06-01
Nerve/centrelines

elias-soltani:
I would like to add a scaffold for nerve centrelines.


elias-soltani:
This is no longer required. The scaffold with the centrelines is published. It could be useful to automatically generate the centrelines from the flatmap. The flatmap has the connectivities between the markers, but we need to give them coordinates. I will create another PR if this becomes more useful.

PR #201, opened 2022-07-28
New stomach

mlin865:


rchristie (changes requested):
I'm super happy with the stomach, but I can't pull this PR in with what appear to be several other changes which aren't part of the stomach: hip, femoralhead, colon motility, Brookes terms etc. and other changes. Also, the new stomach is in Stomach1 but there is also a new Stomach2? Are you intending to replace the previous Stomach1 (I think so)? Only one or other should be modified/added.
I suggest starting a new branch and just putting in your new version only, with minimal changes to anything else. Start with the current main branch.
One minor fix: can you rename the parameter set "Stomach 1" --> "Material" for consistency with the lung scaffold.
If I am slow to get it reviewed, can you add the documentation following the style of the lung.rst file.

rchristie:
Closing as new PR #203 opened with just stomach changes.

PR #209, opened 2022-08-17
Add bladder markers

elias-soltani:
Markers for Keast nerves

### DIFF
--- a/src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
+++ b/src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
@@ -8,7 +8,8 @@ import copy
 import math
 
 from opencmiss.utils.zinc.field import findOrCreateFieldGroup, findOrCreateFieldNodeGroup, \
-    findOrCreateFieldStoredMeshLocation, findOrCreateFieldStoredString
+    findOrCreateFieldStoredMeshLocation, findOrCreateFieldStoredString, findOrCreateFieldCoordinates
+from opencmiss.utils.zinc.finiteelement import getMaximumNodeIdentifier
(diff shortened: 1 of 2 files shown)


src/scaffoldmaker/meshtypes/meshtype_3d_bladderurethra1.py
```diff
@@ -1163,6 +1164,25 @@ def generateBaseMesh(cls, region, options):
...
+            "fundus of urinary bladder": [14.6, 0.01, 23.1],
+            "external urethral sphincter": [-1.5, 6.7, 110.5],
+        }
+        bladder_coordinates = findOrCreateFieldCoordinates(fm)
+        nodeIdentifier = max(1, getMaximumNodeIdentifier(nodes) + 1)
+        for termName, brainstemCoordinatesValues in markerTermNameBrainstemCoordinatesMap.items():
+            annotationGroup = findOrCreateAnnotationGroupForTerm(annotationGroups, region, get_bladder_term(termName))
+            annotationGroup.createMarkerNode(nodeIdentifier, bladder_coordinates, brainstemCoordinatesValues)
+            nodeIdentifier += 1
+
```
rchristie, line 1185: Firstly, this would be OK if bladder_coordinates exist, but they do not.
(And it would need to use "bladder" rather than "brainstem" so the latter word should not appear anyway.)
Currently this will only work for human and when no-one changes any central path or size parameters, since coordinates are being used as material coordinates.
We need to discuss what's best here; I think it's better to use only element & xi until actual material coordinates exist. The above code won't work for other species - the locations will be different/wrong for other species.

rchristie (changes requested):
We need to discuss the comment and how to fix things.

PR #213, opened 2022-10-05
Unit stomach

mlin865:
Hi Richard, 

I've scaled stomach scaffolds for all species to have a geometric central path of 2.0 unit and the scaffolds (material coordinates too) are aligned such that the junction between fundus and body along the central path now sits on the origin. The markers' material coordinates are updated to reflect the changes in alignment of the material coordinates. Unit tests are also updated accordingly. Please review. Thanks

Mabelle


mlin865:
Hi Richard,

I've undo the translation of material coordinates and have now translated stomach scaffold for all species such that their fundus-body intersection along their central path align with the corresponding point for material coordinates. Please review. Thanks

src/scaffoldmaker/meshtypes/meshtype_3d_stomach1.py
```diff
@@ -1222,11 +1222,12 @@ def createStomachMesh3d(region, fm, coordinates, stomachTermsAlong, allAnnotatio
     elementsThroughEsophagusWall = GEJSettings['Number of elements through wall']
     ostiumDiameter = GEJSettings['Ostium diameter']
     limitingRidge = options['Limiting ridge']
+    wallThickness = options['Wall thickness'] * GEJSettings['Unit scale']
```
rchristie, line 1225: It's unexpected that the wall thickness is multiplied by the unit scale of the GEJ. I suggest having the actual thickness in unit scale for the wall thickness, and dividing by GEJSettings['Unit scale'] when transferring to the GEJ in updateSubScaffoldOptions.
I hadn't thought about the GEJ options all needing rescalin , sorry. I think it's OK if they keep their custom unit scale (as you've done) as long as the parent wall thickness is independent as above.
(Also, in hindsight I would have aligned not just the fundus-body junction on the central path, but also it's direction at that point. Not asking you to do this as we've bigger fish to fry.)

rchristie (changes requested):
Small change as noted.

mlin865:
Change made, please review. Thanks

rchristie (approved):
Great.

PR #216, opened 2022-11-02
Marker point annotations do not have their own group

### LINKED ISSUE #215: Marker node annotations should not make zinc groups
We're getting increasing numbers of marker point annotations, e.g. almost 500 for the nerves in the body which makes it impractical to build zinc group fields (which themselves have a group and a node group) for each of them.

Proposal is to make these as annotation groups without an associated zinc group. As now, they are all put in the "marker" group and have a "marker_name" field. Change should not affect anything in the mapping tools.

rchristie:
Fixes #215.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ source_license = readfile("LICENSE")
 
 setup(
     name="scaffoldmaker",
-    version="0.7.1",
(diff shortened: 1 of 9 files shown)


mlin865 (approved):
Looks good

PR #222, opened 2023-04-19
Decode JSON containing sub-scaffoldPackages

Kayvv:


hsorby (approved):
Nice, @rchristie  for visibility.

rchristie (approved):
Looks good, we will have more than one nested Scaffold package in some cases.

PR #231, opened 2023-07-06
Modified bladder

zekh167:
Hi Richard. I fixed the issues you pointed.


rchristie (approved):
What has been done is an improvement for rat, human and code formatting.
However it hasn't made unit-scale central paths which will I have prepared in another PR #233.

PR #232, opened 2023-07-13
Diverging bifurcation

zekh167:
I added two different files creating 2D and 3D diverging bifurcation tubes and changed the tubemesh file to get the right annotations.


rchristie:
Closing as being replaced with and resolved by uterus 2, and generic bifurcation improvments.

PR #234, opened 2023-07-21
Cecum new

mlin865:
Hi Richard,

Note that Ostium 2 only works for 1 vessel at this stage. I kept the code for making multiple vessels (from Ostium 1) in Ostium 2 but they are commented out at this stage. I am hoping to use them later when I start working on implementing multiple vessels for Ostium 2 later on. 

Rest of the PR mainly covers the new cecum scaffold and I switched esophagus, small intestine and colon scaffolds to use 1D network layout instead of 1D path. For stomach, I switched over to the 1D network layout to describe the path from fundus to duodenum but I have yet to implement the bifurcation path that joins the esophagus to the stomach. I will work on that after we have a good version of stomach scaffold. Please review.

Thanks,
Mabelle


mlin865:
@rchristie - I've made all the changes and added the gastrointestinal tract scaffold as well. This PR is now ready for your review.

tests/test_colon.py
```diff
@@ -29,14 +29,10 @@ def test_colon1(self):
         self.assertEqual(parameterSetNames, ["Default", "Cattle 1", "Human 1", "Human 2", "Human 3", "Mouse 1",
                                              "Mouse 2", "Pig 1"])
 
-        centralPathDefaultScaffoldPackages = {
-            'Test line': ScaffoldPackage(MeshType_1d_path1, {
+        parameterSetStructureStrings = {
```
rchristie, line 32: Here and in test small intestine.
Can you rename this and similar variables which aren't structure strings.
Similarly where used below
centralPathOption = parameterSetStructureStrings['Test line']
and
'Central path': copy.deepcopy(centralPathOption)
(surely the above should now be 'Network layout'?)
Rather than copy.copy, best is to construct the ScaffoldPackage in the above line with the dictionary looked up for 'Test line'.
e.g. 
'Network layout': ScaffoldPackage(MeshType_1d_network_layout1, networkLayoutScaffoldPackageDct['Test line'])
There is also a bit of test code with variables centralPath which should be networkLayout.

rchristie (changes requested):
You've already fixed the number of elements through wall.
A couple of additional fixes, which should be simple.

rchristie (approved):
Fantastic scaffold!

PR #235, opened 2023-07-26
Bladder coordinates

zekh167:
Hi Richard. I made the following changes based on our last discussion: 

1. Changed the Material central path's coordinates and derivatives based on the magnitudes obtained from an ellipsoid with height 2.0 and diameter 1.0.
2. Changed the nodes to be made from outer surface toward the inner surface. The only thing I was thinking about is the name of "getCoordinatesFromInner" function in tubemesh.py which I suggest to be "getCoordinatesOffset" or "getTubeCoordinatesOffset" or something else since this function can be used to create both the "outer from inner" and the "inner from outer".
3. Changed the marker locations to put them on the inner surface as before.
4. Changed the wall thicknesses for different species.
5. Adjusted the test_bladder.py code.


src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -346,25 +352,25 @@ def getDefaultOptions(cls, parameterSetName='Default'):
...
-            options['Wall thickness'] = 0.003  # was 0.3 / 110.40996641878101
+            options['Wall thickness'] = 0.006  # was 0.3 * 2.0 / 110.40996641878101
             options['Ureter position around'] = 0.67  # should be on the dorsal part (> 0.5)
         if 'Material' in parameterSetName:
             options['Number of elements along dome'] = 8
             options['Number of elements along neck'] = 4
             options['Number of elements around'] = 8
-            options['Wall thickness'] = 0.01
+            options['Wall thickness'] = 0.016  # an average of the wall thicknesses for above species
             options['Ureter position around'] = 0.67
```
rchristie, line 374: Since we require the 'Ureter position around' to be the same for all variants, remove it from the per-species and material code. Add a comment explaining that it's a fixed material coordinate and thus not to be edited.
Remove this option from the visible options in getOrderedOptionNames() below.

rchristie (changes requested):
I noticed serosa/lumen annotations were reversed, but the problem is caused by incorrect extrusion from outer surface (use Hermite through wall to see D3 is inward; I need to see why the element axes are not showing the problem). See comment in code for fixing - should be straightforward.
See also comment about hiding fixed ureter position option.
Please be very careful to check annotations and tests are fine.
Other than that I'm very happy with the material coordinates.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -1297,7 +1335,7 @@ def getBladderCoordinates(elementsCountAlongDome, elementsCountAlongNeck, elemen
             d3Inner.append(d3)
 
     # Create outer layers from the inner nodes
-    wallThicknessList = [wallThickness] * (elementsCountAlongBladder + 1)
+    wallThicknessList = [-wallThickness] * (elementsCountAlongBladder + 1)
     relativeThicknessList = []
     transitElementList = [0] * elementsCountAround
     xList, d1List, d2List, d3List, curvatureList = \
```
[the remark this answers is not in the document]
zekh167: - I renamed Mabelle's tubemesh function and added the outward parameter to this function in all previous uses. Nothing else has been changed for the other scaffolds' code. 
- I also changed all variables in that function in tubemesh and documented the function arguments.
- Changed the tubemesh function in order to create the nodes order from inner to outer, d3 pointing from inside to out while the bladder scaffold is created inward (outward=False). It also works properly for the outward=True in other scaffolds like colonsegment and esophagus.
- Adjusted the markers location.
- Checked the annotation to be true.
- Bladder test has been updated. All tests pass.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
zekh167: Done as suggested above.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -614,14 +608,14 @@ def generateBaseMesh(cls, region, options):
                                                               get_bladder_term("right ureter junction with bladder"))
 
         idx1 = 1
-        xi1 = [0.0, 0.0, 1.0]
+        xi1 = [0.0, 0.0, 0.0]
         markerList = []
         markerList.append({"group": apexGroup, "elementId": idx1, "xi": xi1})
         idx2 = elementsCountAlongDome * elementsCountAround * elementsCountThroughWall + ureterElementPositionAround + 1
-        xi2 = [ureter1Position.xi1, 0.0, 1.0]
+        xi2 = [ureter1Position.xi1, 0.0, 0.0]
```
rchristie, line 615: This should have remained at 1.0 ? Want apex marker on the outside. Check it is correct and the bladder test still passes!

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -344,7 +344,7 @@ def getDefaultOptions(cls, parameterSetName='Default'):
             'Number of elements around': 8,
             'Number of elements through wall': 1,
             'Wall thickness': 0.016,
```
rchristie, line 346: Default parameters use the cat central path but the material wall thickness - please change to make cat parameters the default. Just need to change wall thickness to 0.007 above.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
```diff
@@ -353,25 +353,19 @@ def getDefaultOptions(cls, parameterSetName='Default'):
...
             options['Wall thickness'] = 0.016  # was 2.5 * 2.0 / 505.2045520616655
-            options['Ureter position around'] = 0.67  # should be on the dorsal part (> 0.5)
         if 'Rat 1' in parameterSetName:
             options['Wall thickness'] = 0.006  # was 0.3 * 2.0 / 110.40996641878101
-            options['Ureter position around'] = 0.67  # should be on the dorsal part (> 0.5)
         if 'Material' in parameterSetName:
             options['Number of elements along dome'] = 8
             options['Number of elements along neck'] = 4
             options['Number of elements around'] = 8
             options['Wall thickness'] = 0.016  # an average of the wall thicknesses for above species
```
rchristie, line 368: Make the wall thickness in material coordinates 0.02 so it's a nice round number (and closer to half the human wall thickness). I'm still surprised by the variations here...
```
            options['Wall thickness'] = 0.02  # an average of the wall thicknesses for above species, rounded
```

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -243,28 +243,27 @@ def warpSegmentPoints(xList, d1List, d2List, segmentAxis, sx, sd1, sd2, elements
...
-    :param xInner: Coordinates on inner surface
-    :param d1Inner: Derivatives on inner surface around tube
-    :param d2Inner: Derivatives on inner surface along tube
-    :param d3Inner: Derivatives on inner surface through wall
+    Generates extruded coordinates using coordinates and derivatives of a surface.
+    :param xSurf: Coordinates on surface
+    :param d1Surf: Derivatives on surface around tube
+    :param d2Surf: Derivatives on surface along tube
+    :param d3Surf: Derivatives on surface through wall
     :param wallThicknessList: Wall thickness for each element along tube
```
rchristie, line 255: Is this actually for each *node* along tube?

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -243,28 +243,27 @@ def warpSegmentPoints(xList, d1List, d2List, segmentAxis, sx, sd1, sd2, elements
...
     :param relativeThicknessList: Relative wall thickness for each element through wall
     :param elementsCountAround: Number of elements around tube
     :param elementsCountAlong: Number of elements along tube
     :param elementsCountThroughWall: Number of elements through tube wall
     :param transitElementList: stores true if element around is a transition
-    element that is between a big and a small element.
```
rchristie, line 262: Restore this comment line.

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -281,71 +280,87 @@ def getCoordinatesFromInner(xInner, d1Inner, d2Inner, d3Inner,
         relativeThicknessList.append(relativeThicknessList[-1])
 
     for n2 in range(elementsCountAlong + 1):
-        wallThickness = wallThicknessList[n2]
+        if outward:
+            wallThickness = wallThicknessList[n2]
+        else:
+            wallThickness = -wallThicknessList[n2]
```
rchristie, line 286: The code below has several near identical cases for inward and outward which are different only because of the above sign. Don't modify the wallThickness, instead have a new variable:
```
wallOutwardDisplacement = wallThickness if outward else -wallThickness
```
Use it in the extrusion code below (see other comment).

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -281,71 +280,87 @@ def getCoordinatesFromInner(xInner, d1Inner, d2Inner, d3Inner,
...
         for n1 in range(elementsCountAround):
             n = n2*elementsCountAround + n1
-            norm = d3Inner[n]
-            # Calculate outer coordinates
-            x = [xInner[n][i] + norm[i]*wallThickness for i in range(3)]
-            xOuter.append(x)
+            norm = d3Surf[n]
+            # Calculate extruded coordinates
+            x = [xSurf[n][i] + norm[i]*wallThickness for i in range(3)]
```
rchristie, line 291: Use the signed wall displacement here only:
```
            x = [xSurf[n][i] + norm[i]*wallOutwardDisplacement for i in range(3)]
```

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -281,71 +280,87 @@ def getCoordinatesFromInner(xInner, d1Inner, d2Inner, d3Inner,
...
+                surfx = xSurf[n]
+                extrudedx = xExtrudedSurf[n]
                 # x
-                x = interp.interpolateCubicHermite(innerx, dWall, outerx, dWall, xi3)
+                if outward:
+                    dWall = [wallThickness * c for c in norm]
+                    x = interp.interpolateCubicHermite(surfx, dWall, extrudedx, dWall, xi3)
+                else:
+                    dWall = [-wallThickness * c for c in norm]
+                    x = interp.interpolateCubicHermite(extrudedx, dWall, surfx, dWall, xi3)
```
rchristie, line 337: Still need separate cases for x = ~, but dWall = should not depend on inward/outward.

rchristie (changes requested):
Some minor fixes and code improvements.
Please eyeball changes and run tests again after changing.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
zekh167: I moved the apex to outside surface and updated the test. Done.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
zekh167: Done.

src/scaffoldmaker/meshtypes/meshtype_3d_bladder1.py
zekh167: Done.

src/scaffoldmaker/utils/tubemesh.py
zekh167: Restored.

src/scaffoldmaker/utils/tubemesh.py
zekh167: Done.

src/scaffoldmaker/utils/tubemesh.py
zekh167: Done.

src/scaffoldmaker/utils/tubemesh.py
zekh167: Done. Kept the dWall the same, but different x.

zekh167:
All above comments applied and updated the bladder, bladderurethra, cecum, colon, colonsegment, esophagus and smallintestine tests.

src/scaffoldmaker/utils/tubemesh.py
zekh167: I think this represents the wallthickness for each layer of elements along tube.

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -282,70 +282,77 @@ def getCoordinatesFromInner(xInner, d1Inner, d2Inner, d3Inner,
...
+                if outward:
+                    x = interp.interpolateCubicHermite(surfx, dWall, extrudedx, dWall, xi3)
+                else:
+                    x = interp.interpolateCubicHermite(extrudedx, dWall, surfx, dWall, xi3)
                 xList.append(x)
 
                 # dx_ds1
-                factor = 1.0 + wallThickness*xi3 * curvatureAroundInner[n]
-                d1 = [ factor*c for c in d1Inner[n]]
+                factor = 1.0 - wallOutwardDisplacement * xi3 * curvatureAroundSurf[n]
```
rchristie, line 339: The old code was correct (except for rename of Inner -> Surf):
```
                factor = 1.0 + wallThickness * xi3 * curvatureAroundSurf[n]
```

rchristie (changes requested):
d1 calculation is incorrect in extrusion.
Fix comment.
cecum, colon, colon segment, esophagus and small intestine test answers should not have changed - please restore them.

src/scaffoldmaker/utils/tubemesh.py
```diff
@@ -282,70 +282,77 @@ def getCoordinatesFromInner(xInner, d1Inner, d2Inner, d3Inner,
...
+                    factor = 1.0 - curvature * distance
+                else:
+                    factor = 1.0 + curvature * distance
+                d2 = [factor*c for c in d2Surf[n]]
                 d2List.append(d2)
                 curvatureList.append(curvature)
 
                 #dx_ds3
-                d3 = [c * wallThickness * (relativeThicknessList[n3] if relativeThicknessList else 1.0/elementsCountThroughWall) for c in norm]
+                d3 = [c * -wallOutwardDisplacement * (relativeThicknessList[n3] if relativeThicknessList else 1.0 / elementsCountThroughWall) for c in norm]
```
rchristie, line 355: Old code was fine. Use wallThickness not -wallOutwardDisplacement which was reversing colon d3.

rchristie (changes requested):
Fix d3 in tubemesh

rchristie (approved):
Looking great now.

PR #237, opened 2023-08-17
Handle d1 scaling when using inward and update unit test for bladder

mlin865:


rchristie (approved):
Approve as it's now consistent.
I'm slightly concerned that we're subtracting the curvature to get the factor; makes me wonder why it's getting a negative value for the outward case.
Also noting that relativeThicknessList will need to be used in reverse order for inward extrusion.

PR #238, opened 2023-08-31
Test uterus network

zekh167:
Hi, this is the mouse and rat uterus scaffold with different number of elements around horns and body and also through the wall. The scaffold was created inward. So, the network layout parameters control the outer shape.


rchristie (approved):
The code runs well and the model looks good.
I note the following issues with the code:
- Uterus terms are incomplete, many without ontology IDs, or with names which are not specific enough e.g. "serosa of right horn" should be "serosa of right uterine horn" or similar.
- No test code for uterus.
- No material coordinates
These will need to be fixed at a later data.

PR #241, opened 2023-11-29
Smooth side derivatives and inner coordinates

rchristie:
Support inner coordinates in all network layout interactive functions.

Implement smooth side cross derivatives correctly per segment chain, handling different versions.
If there is a selection group, only smooths segment chains containing a selected element.


mlin865 (approved):
Works well and passed all unit tests.

PR #247, opened 2024-03-26
Limit bifurcation trim surface intersections to half curve

rchristie:
Fix case where two segments are intersecting at both ends, ensuring only the regions near the respective end are sampled for initial nearest curve location when finding curve/track surfaces intersections used for building bifurcation trim surfaces.


mlin865 (approved):
Looks good, I tested it with existing networks and also made something like the pelvic bone  and they work well. Passed all the tests. Only uterus 1 isn't working (not related to your changes) and it is a fix I addressed in my uterus pull request so once that is merged, uterus 1 will work fine.

PR #252, opened 2024-04-22
Avoid testing equality of Field and None

rchristie:


hsorby (approved):
Awesome, good to see those green ticks.

PR #261, opened 2024-07-23
Remove matrix utility file use cmlibs.maths 0.6.2 instead.

hsorby:


tests/test_gastrointestinaltract1.py
```diff
@@ -113,7 +113,7 @@ def test_gastrointestinaltract1(self):
         self.assertAlmostEqual(surfaceArea, 280350.6462177311, delta=1.0E-6)
         result, volume = volumeField.evaluateReal(fieldcache, 1)
         self.assertEqual(result, RESULT_OK)
-        self.assertAlmostEqual(volume, 598505.4701647629, delta=1.0E-6)
+        self.assertAlmostEqual(volume, 598505.4701647585, delta=1.0E-6)
```
rchristie, line 116: That delta is way too low for the magnitude of the value; better to set it to e.g. 1.0E-3.

rchristie (approved):
One minor fix in comment, otherwise approved.

tests/test_gastrointestinaltract1.py
hsorby: [a61189c](https://github.com/ABI-Software/scaffoldmaker/pull/261/commits/a61189c2cbc0e6e7d31b953f7c2ddee03f28ef14)

PR #268, opened 2024-11-22
New cap mesh

oneway1225:
Add cap mesh that generates cap(s) at either ends of a tube network.
Works by adding "(" and ")" at the ends of a network layout structure, e.g. (1-2) creates a default tube network with caps at both ends.


oneway1225:
kidney scaffoldmaker1
kidney scaffoldmaker2

- Added a new bend function.
- 1D kidney network layout is set to a private scaffold type. The network layout edit will be only used within the 3D kidney scaffold and be used for adjusting the dimensions of the kidney.

PR #269, opened 2024-12-11
Trigeminal nerve and spinal nerve scaffolds

mlin865:


src/scaffoldmaker/meshtypes/meshtype_3d_spinalnerve1.py
```diff
@@ -0,0 +1,705 @@
...
+from cmlibs.utils.zinc.field import Field, find_or_create_field_coordinates
+from cmlibs.zinc.node import Node
+from scaffoldmaker.annotation.annotationgroup import AnnotationGroup
+from scaffoldmaker.annotation.spinal_nerve_terms import get_spinal_nerve_term
+from scaffoldmaker.meshtypes.meshtype_1d_network_layout1 import MeshType_1d_network_layout1
+from scaffoldmaker.meshtypes.scaffold_base import Scaffold_base
+from scaffoldmaker.scaffoldpackage import ScaffoldPackage
+from scaffoldmaker.utils.interpolation import sampleCubicHermiteCurves, smoothCubicHermiteDerivativesLine
+from scaffoldmaker.utils.networkmesh import NetworkMesh
+from scaffoldmaker.utils.tubenetworkmesh import BodyTubeNetworkMeshBuilder, TubeNetworkMeshGenerateData
```
rchristie, line 15: Groups dorsal/ventral/left/right aren't correct.
Could remove; I think you only need to derive from TubeNetworkMeshBuilder to do this.
However, when you have the human pair or the whole spine it would be nice to label left and right.
Not so sure dorsal-ventral are worth keeping in their current form as there are already meaningful groups with those names for the nerve.

src/scaffoldmaker/meshtypes/meshtype_3d_trigeminalnerve1.py
```diff
@@ -0,0 +1,591 @@
...
+    def getDefaultOptions(cls, parameterSetName="Default"):
+        options = {}
+        useParameterSetName = "Human 1" if (parameterSetName == "Default") else parameterSetName
+        options["Base parameter set"] = useParameterSetName
+        options["Trigeminal nerve network layout"] = ScaffoldPackage(MeshType_1d_human_trigeminal_nerve_network_layout1)
+        options["Number of elements along nerve root"] = 5
+        options["Number of elements along trigeminal ganglion"] = 3
+        options["Number of elements along mandibular nerve"] = 4
+        options["Number of elements along maxillary nerve"] = 4
+        options["Number of elements along ophthalmic nerve"] = 5
```
rchristie, line 388: Why so many elements along? 3,2,3,3,3 looks acceptable.

src/scaffoldmaker/meshtypes/meshtype_3d_trigeminalnerve1.py
```diff
@@ -0,0 +1,591 @@
...
+from cmlibs.zinc.node import Node
+from scaffoldmaker.annotation.annotationgroup import AnnotationGroup
+from scaffoldmaker.annotation.trigeminal_nerve_terms import get_trigeminal_nerve_term
+from scaffoldmaker.meshtypes.meshtype_1d_network_layout1 import MeshType_1d_network_layout1
+from scaffoldmaker.meshtypes.scaffold_base import Scaffold_base
+from scaffoldmaker.scaffoldpackage import ScaffoldPackage
+from scaffoldmaker.utils.interpolation import (getCubicHermiteArcLength, interpolateLagrangeHermiteDerivative,
+                                               sampleCubicHermiteCurvesSmooth)
+from scaffoldmaker.utils.networkmesh import NetworkMesh
+from scaffoldmaker.utils.tubenetworkmesh import BodyTubeNetworkMeshBuilder, TubeNetworkMeshGenerateData
```
rchristie, line 15: Groups dorsal/ventral/left/right don't make any sense. I think you only need to derive from TubeNetworkMeshBuilder to remove them. [I don't want to do this at the moment, but technically dorsal/ventral could become left/right or medial lateral; the latter are painful as you'd need to define it it's left or right trigeminal.]

src/scaffoldmaker/meshtypes/meshtype_3d_spinalnerve1.py
```diff
@@ -0,0 +1,705 @@
...
+    def getDefaultOptions(cls, parameterSetName="Default"):
+        options = {}
+        useParameterSetName = "Human 1" if (parameterSetName == "Default") else parameterSetName
+        options["Base parameter set"] = useParameterSetName
+        options["Spinal nerve network layout"] = ScaffoldPackage(MeshType_1d_human_spinal_nerve_network_layout1,
+                                                                 defaultParameterSetName=useParameterSetName)
+        options["Number of elements along nerve root"] = 5
+        options["Number of elements along dorsal root around DRG"] = 2
+        options["Number of elements along dorsal root ganglion"] = 4
+        options["Number of elements along ventral root"] = 8
```
rchristie, line 506: Probably too many elements, what about 5,2,4,8 -> 3,1,3,5.
Rename "along dorsal root around DRG" to "along dorsal root beside DRG", as "around" has a special meaning.

rchristie (changes requested):
The whole spine takes a long time to make. Reducing the number of elements as noted should help.
Can you change to human spine in the network layout from any other parent parameters set? I didn't wait for this.
I note that we should allow zero elements in the shell as they're needless here.
I would normally ask for a test, but as with the point about the shell, we have another plan for these scaffolds in future.

mlin865:
Hi Richard, all requested changes have been made and ready for your review.

rchristie (changes requested):
One small fix needed. Please test left/right on Human pair 1.

PR #273, opened 2025-03-21
Fit trunk radius and orientation

rchristie:


docs/scaffolds/vagus.rst
v-chopovda, line 1: doc needs updating (we can do this later)
* add straight coordinates
* update input file requirements (update approved landmarks, trunk group names)

PR #277, opened 2025-07-17
Vagus metadata

rchristie:
Add support for getting additional metadata out of scaffolds.


rchristie:
Requires Requires PR ABI-Software/fieldfitter#11.

PR #286, opened 2025-10-22
Adding joint movements into the wholebody scaffold

0huitzil:
Continuous work regarding the addition of joint flexion and abduction movements into the wholebody scaffold. 
Highlights
- Network layout structure for the 1d stickman is now created using a script, using the number of elements on each section as parameters. Both this script and the number of elements for each body section are now stored on a new file utils/human_network_layout.py 
- Arm has been divided into upper arm (brachium) and lower arm (antebrachium), both on the 1D and 3D scaffolds 
- Leg has been divided into upper leg and lower leg, both on the 1D and 3D scaffolds 
- Added flexion functionality at the elbows, knees and ankles. Flexion angle is a user-modifiable parameter, which changes the position and frame of reference at the joint node to ensure a smooth transition in the tube network. This code is a temporary fix, which works well for all the flexion angles tested, but causes deformations in the scaffold in the number of elements on the corresponding region of the 3D scaffold is changes. From my testing, it seems like there is no way to avoid this unless we set two frames of reference at the joint node (which is not currently possible). But I could be wrong. 
- Some flexion functionality has been added into the shoulder joint as well, but this is still a work in progress. The current code does not work well for flexion angles > 90 degrees. Most likely, the same mechanism for adding flexion into the shoulder will work for the hip (which is the other mayor flexion movement missing) 
- Also fixed a missing import on the split function in utils/networkmesh. I ended up not needing this fix, but I am including it still as a bugfix.


ghost:
@0huitzil do these tests need to pass before RC reviews it? 
What's the status of things (from memory there should be a PR)?

0huitzil:
@dananana-batman This is the PR for the joints itself. The tests are very sensitive to small changes in the scaffold, which I have to adjust every time I push changes to it. For the sake of time, I only update them semi-regularly. Though I will make sure to update them again once RC is back and has time to look at the PR

PR #287, opened 2025-12-01
Add lung4 scaffold with open fissures

rchristie:
Built from ellipsoid octants, with 1 row of nodes on inner hilum common to all lobes on each lung.

### DIFF
--- a/src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
+++ b/src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
@@ -463,17 +463,6 @@ class MeshType_3d_lung4(Scaffold_base):
                 for i in range(4 if (lung == left_lung) else 5):
                     marker_name_element_xi.append(("dummy", None, None))
 
-        # marker points; make after regular nodes so higher node numbers
-
-        lung_nodeset = lung_group.getNodesetGroup(nodes)
-        for marker_name, element, xi in marker_name_element_xi:
-            if element:  # skip dummy markers for missing left/right, but increment node_identifier always
-                annotation_group = findOrCreateAnnotationGroupForTerm(
-                    annotation_groups, region, get_lung_term(marker_name), isMarker=True)
-                marker_node = annotation_group.createMarkerNode(node_identifier, element=element, xi=xi)
-                lung_nodeset.addNode(marker_node)
-            node_identifier += 1
-
         for lung in lungs:
             is_left = lung == left_lung
             lung_nodeset = (left_lung_group if is_left else right_lung_group).getNodesetGroup(nodes)
@@ -493,6 +482,20 @@ class MeshType_3d_lung4(Scaffold_base):
 
             translate_nodeset_coordinates(lung_nodeset, coordinates,
                                           [-lung_spacing if is_left else lung_spacing, 0.0, 0.0])
+
+        # marker points; make after regular nodes so higher node numbers
+        lung_nodeset = lung_group.getNodesetGroup(nodes)
+        for marker_name, element, xi in marker_name_element_xi:
+            if element:  # skip dummy markers for missing left/right, but increment node_identifier always
+                annotation_group = findOrCreateAnnotationGroupForTerm(
+                    annotation_groups, region, get_lung_term(marker_name), isMarker=True)
+                marker_node = annotation_group.createMarkerNode(node_identifier, element=element, xi=xi)
+                if 'left' in marker_name:
+                    left_lung_group.getNodesetGroup(nodes).addNode(marker_node)
+                else:
+                    right_lung_group.getNodesetGroup(nodes).addNode(marker_node)
+                lung_nodeset.addNode(marker_node)
+            node_identifier += 1
 
         return annotation_groups, None
 
--- a/src/scaffoldmaker/utils/meshrefinement.py
+++ b/src/scaffoldmaker/utils/meshrefinement.py
@@ -170,7 +170,7 @@ class MeshRefinement:
                     tmp_line_ids.append(line.getIdentifier())
             face_line_ids.append(tmp_line_ids)
         new_line_ids = set()
-        # if there is a single line between 2 faces can do less work later, but not if there are collpased faces
+        # if there is a single line between 2 faces can do less work later, but not if there are collapsed faces
         single_line = initial_face_count < 3
         for f1 in range(len(faces) - 1):
             for f2 in range(f1 + 1, len(faces)):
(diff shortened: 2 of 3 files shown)


src/scaffoldmaker/meshtypes/meshtype_3d_heartarterialroot1.py
```diff
@@ -129,7 +129,7 @@ def generateBaseMesh(cls, region, options, baseCentre=[ 0.0, 0.0, 0.0 ], axisSid
             #     AnnotationGroup(region, get_heart_term("posterior cusp of aortic valve")),
             #     AnnotationGroup(region, get_heart_term("right cusp of aortic valve")),
             #     AnnotationGroup(region, get_heart_term("left cusp of aortic valve")) ]
-            cuspGroup = AnnotationGroup(region, get_heart_term("aortic valve leaflet")),
+            cuspGroup = AnnotationGroup(region, get_heart_term("aortic valve leaflet"))
```
mlin865: I noticed that there are 3 entries of "aortic valve leaflet" in Annotation Tools and selection of any of these entries will highlight all 3 leaflets on the scaffold. Is this expected?

src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
```diff
@@ -0,0 +1,804 @@
...
+        return options
+
+    @classmethod
+    def getOrderedOptionNames(cls):
+        return [
+            "Left lung",
+            "Right lung",
+            "Number of elements lateral",
+            "Number of elements lower lobe extension",
+            "Number of elements oblique",
```
mlin865: Would it be clearer if you rename this as 'Number of elements along oblique fissure"?

src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
```diff
@@ -0,0 +1,804 @@
...
+        for key in [
+            "Lower lobe base concavity",
+            "Lower lobe extension",
+            "Cardiac curvature"
+        ]:
+            if options[key] < 0.0:
+                options[key] = 0.0
+        if options["Lower lobe extension"] == 0.0:
+            options["Number of elements lower lobe extension"] = 0
+            dependent_changes = True
```
mlin865: Please check as it is showing a broken mesh at the moment

src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
```diff
@@ -0,0 +1,804 @@
...
+        depth = options["Ellipsoid dorsal-ventral size"]
+        height = options["Ellipsoid height"]
+        max_extension = 0.99 * magnitude(getEllipsePointAtTrueAngle(depth / 2.0, height / 2.0, math.pi / 3.0))
```
mlin865: returns a long string of numbers 0.4634829428966771. Consider truncating /  rounding to less decimal places?

src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
```diff
@@ -0,0 +1,804 @@
...
+    displacement = fieldmodule.createFieldConcatenate([delta_x, delta_y, delta_z])
+    new_coordinates = coordinates + displacement
+    fieldassignment = coordinates.createFieldassignment(new_coordinates)
+    fieldassignment.setNodeset(nodeset)
+    fieldassignment.assign()
+
+
+def taper_lung_edge(sharpeningFactor, fieldmodule, coordinates, nodeset, halfValue, isBase=False):
+    """
+    Applies a tapering transformation to the lung geometry to sharpen the anterior edge or the base.
```
mlin865: Change "Applies" to "Apply"

src/scaffoldmaker/utils/ellipsoidmesh.py
```diff
@@ -314,41 +220,85 @@ def _build_ellipsoid_octant(self, half_counts, axis2_x_rotation_radians, axis3_x
                 start_weight, end_weight, overweighting, end_transition))
         move_x_to_ellipsoid_surface = lambda x: moveCoordinatesToEllipsoidSurface(self._a, self._b, self._c, x)
         move_d_to_ellipsoid_surface = lambda x, d: moveDerivativeToEllipsoidSurface(self._a, self._b, self._c, x, d)
+        def evaluate_surface_d3_ellipsoid_plane(tx, td1, td2):
+            """
+            Restrict d3 to be ellipsoid normal constrained be in plane interpolated from axis_d2 to axis_d3 at tx
+            relative to ext_origin.
+            :return: d3 with magnitude dir_mag.
```
mlin865: Define tx, td1, td2.

src/scaffoldmaker/utils/ellipsoidmesh.py
```diff
@@ -314,41 +220,85 @@ def _build_ellipsoid_octant(self, half_counts, axis2_x_rotation_radians, axis3_x
                 start_weight, end_weight, overweighting, end_transition))
         move_x_to_ellipsoid_surface = lambda x: moveCoordinatesToEllipsoidSurface(self._a, self._b, self._c, x)
         move_d_to_ellipsoid_surface = lambda x, d: moveDerivativeToEllipsoidSurface(self._a, self._b, self._c, x, d)
+        def evaluate_surface_d3_ellipsoid_plane(tx, td1, td2):
+            """
+            Restrict d3 to be ellipsoid normal constrained be in plane interpolated from axis_d2 to axis_d3 at tx
```
mlin865: Seems to be missing a "to"- "Restrict d3 to be ellipsoid normal constrained **to** be in plane interpolated...."

mlin865 (commented):
Scaffold looks great and passed all tests. I've made couple of minor suggestions and also pending your fixes to the two issues discussed today: 
1) issues with edge related annotation groups when using refinement
2) broken mesh when "lower lobe extension" = 0.0

Are you planning to add markers that are in older versions of lung to lung 4?

src/scaffoldmaker/meshtypes/meshtype_3d_lung4.py
rchristie: I think it's the best for now and the oblique plane is understood by lung people. It's not strictly true for the fissure because there are fewer on the lower lobe fissure (half + extension).
Also it may be possible to build the mesh with a different number horizontally, so this leaves that option open (vs. having number of elements radial (half the number) which is not good to mix with the number in the lateral direction (medial-lateral) which I don't want to use half the number for.

rchristie:
> Scaffold looks great and passed all tests. I've made couple of minor suggestions and also pending your fixes to the two issues discussed today:
> 
> 1. issues with edge related annotation groups when using refinement
> 2. broken mesh when "lower lobe extension" = 0.0
> 
> Are you planning to add markers that are in older versions of lung to lung 4?

We're not fixing any marker points -- the apex location is a bit amorphous and everything else is tied down by the surfaces and edges already.

rchristie:
I did add markers which are not used for fitting but in subsequence PCA steps.
Ready for final review.

mlin865 (approved):
Works fine now. Thanks for making the changes

PR #295, opened 2026-07-29
Material coordinates

0huitzil:
Whole-body scaffold is in need of a material coordinates field. This pull request includes the following changes to the wholebody2 scaffold:
- [x] Added a material coordinates field into the scaffold (body coordinates). This is done by calling the tube generator one more time, with default settings. 
- [x] Changes the default proportions of the 1d-network-layout scaffold to be more consistent to those of a "default" human 
- [x] Created a script which automatically generated the network layout of the body (util/human_network_layout)
- [x] Subdivided arms into brachium/antebrachium
- [x] Subdivided legs into upper/lower leg
- [x] Added skin groups for hand and foot


src/scaffoldmaker/scaffolds.py
```diff
@@ -72,6 +93,7 @@ class Scaffolds(object):
     _allScaffoldTypes = [
         MeshType_1d_bifurcationtree1,
         MeshType_1d_uterus_network_layout1,
+        MeshType_1d_human_body_network_layout1,
```
rchristie: This shouldn't be committed to the public scaffold list.
I don't expect any change to this whole file -- can you revert it?

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody2.py
```diff
@@ -1,22 +1,38 @@
...
 from cmlibs.utils.zinc.field import Field, find_or_create_field_coordinates
 from cmlibs.zinc.element import Element
 from cmlibs.zinc.node import Node
+
 from scaffoldmaker.annotation.annotationgroup import (
-    AnnotationGroup, findOrCreateAnnotationGroupForTerm, getAnnotationGroupForTerm)
+    AnnotationGroup,
+    findOrCreateAnnotationGroupForTerm,
+    getAnnotationGroupForTerm,
+)
```
rchristie: Don't needlessly change style like this. I don't want one line per imported function as we can get dozens from some utilities.

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody2.py
```diff
@@ -1,22 +1,38 @@
 """
 Generates a 3D body coordinates using tube network mesh.
 """
+import math
+
```
rchristie: math was already being imported below -- don't change unlesss we agree on a new standard for ordering, and the current approach is that python system libraries are imported last after our own sources.

src/scaffoldmaker/meshtypes/meshtype_3d_wholebody2.py
```diff
@@ -1,22 +1,38 @@
...
+    sampleCubicHermiteCurvesSmooth,
+    smoothCubicHermiteDerivativesLine,
+)
 from scaffoldmaker.utils.networkmesh import NetworkMesh
-from scaffoldmaker.utils.tubenetworkmesh import BodyTubeNetworkMeshBuilder, TubeNetworkMeshGenerateData
-import math
+from scaffoldmaker.utils.tubenetworkmesh import (
+    BodyTubeNetworkMeshBuilder,
+    TubeNetworkMeshGenerateData,
+)
```
rchristie: Revert these unnecessary changes.

src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -10,8 +10,20 @@
     ("abdominopelvic cavity", "UBERON:0035819"),
     ("upper limb", "UBERON:0001460"),
     ("left upper limb", "UBERON:8300002", "FMA:7186"),
+    ("left shoulder", ""),
```
rchristie: Do any of these new terms have ILX or UBERON ids?
Look up on interlex...

src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -21,11 +33,20 @@
     ("head core", ""),
     ("diaphragm", "UBERON:0001103", "ILX:0103194"),
     ("hand", "ILX:0104885", "FMA:9712"),
+    ("hip", ""),
```
rchristie: Does hip have a real anatomical definition? Perhaps needs to be a more focused term for the level of the top of the pelvis?

src/scaffoldmaker/utils/human_network_layout.py
```diff
@@ -0,0 +1,139 @@
+#%%
```
rchristie: I wouldn't have started a new utility that is used in one place, but it's OK.
However, need to have header text explaining what it is for.

src/scaffoldmaker/utils/human_network_layout.py
```diff
@@ -0,0 +1,139 @@
...
+    'handElementsCount': 1,
+    'thoraxElementsCount': 3,
+    'abdomenElementsCount': 4,
+    'hipElementsCount': 2,
+    'upperLegElementsCount': 2,
+    'lowerLegElementsCount': 2,
+    'footElementsCount': 2
+}
+
+def create_segment_layout(nodeCount:int, nodeIdentifier:int,
```
rchristie: This is a very generic name. Is the function able to be used for other scaffolds?
It seems like it should be a method of a class that maintains the next node identifier...

src/scaffoldmaker/utils/human_network_layout.py
```diff
@@ -0,0 +1,139 @@
...
+                        initialJointNode:int = 0, versionStart:int = 0, versionEnd:int = 0):
+    """
+    Construct a segment of the human network node
+
+    :param nodeCount: Number of nodes to add.
+
+    :param networkLayout: String containing the current network layout.
```
rchristie: No such argument. A NetworkLayout is a specific class. This would logically be what we call a structure string.

src/scaffoldmaker/utils/human_network_layout.py
```diff
@@ -0,0 +1,139 @@
+#%%
+# Number of elements per segment. Used to calculate the number of nodes per segment.
+human_network_element_counts = {
```
rchristie: This might be better done as a members of a class?
The generate function would then be a method on that class?

rchristie (changes requested):
Not really a full review as this is a larger change than I expected (which was just to add material coordinates). For a larger change including upper/lower limbs etc. I will need to look at when I'm back from Mon 17.
There are some comments about unnecessary formatting changes, but little real substance due to the following:
For some reason the meshtype_3d_wholebody2 scaffold diff is unable to be rendered in the github interface. Has it caught up to the tube mesh dome changes?
I'll need you to walk me through the changes now they're a bit larger.

0huitzil:
On a second look, the limb subdivision changes are definitely undercooked. And, given the usefulness of the material coordinates to the other projects we are working on, I decided to move the other limb subdivision changes to a separate branch, and just keep the following changes:

- [x] Added material coordinates to the scaffold
- [x] Updated default lengths of the 1d network
- [x] Updated the test to the new default parameters

rchristie (changes requested):
Can we change to the following default sizes in the body coordinates branch:
* "Abdomen length": 3.0,
* "Hand length": 2.0,
* "Head length": 2.5,
* "Neck length": 1.0,
* "Shoulder width": 4.5,
* "Thorax length": 2.5,

We'll tweak this again in subsequent stuff, but I think these will be better.
Can you make these changes, update test answers, commit and remove 'draft' from the PR.

src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -21,13 +21,21 @@
     ("head core", ""),
     ("diaphragm", "UBERON:0001103", "ILX:0103194"),
     ("hand", "ILX:0104885", "FMA:9712"),
+    ("right hand", "ILX:"),
+    ("left hand", "ILX:"),
+    ("right hand skin epidermis outer surface", "ILX:"),
+    ("left hand skin epidermis outer surface", "ILX:"),
```
rchristie: If there is not term ID use `""` (empty quotes)

src/scaffoldmaker/annotation/body_terms.py
```diff
@@ -21,13 +21,21 @@
...
     ("left lower limb", "UBERON:8300004", "FMA:24981"),
     ("left lower limb skin epidermis outer surface", "ILX:0796506"),
-    ("right lower limb ", "UBERON:8300003", "FMA:24980"),
+    ("right lower limb", "UBERON:8300003", "FMA:24980"),
     ("right lower limb skin epidermis outer surface", "ILX:0796505"),
     ("foot", "ILX:0745450", "FMA:9664"),
+    ("right foot", "ILX:"),
+    ("left foot", "ILX:"),
+    ("right foot skin epidermis outer surface", "ILX:"),
+    ("left foot skin epidermis outer surface", "ILX:"),
```
rchristie: Use blank id`""`

rchristie (changes requested):
Minor fixes

